PR #43, opened 2017-04-27
Fix bootstrap script

goruha:
## What 
* Fix bootstrap make target

## Why
* It is required step for cluster create

### DIFF
--- a/rootfs/usr/local/include/toolbox/Makefile
+++ b/rootfs/usr/local/include/toolbox/Makefile
@@ -53,8 +53,8 @@ configure:
 .PHONY : bootstrap
 ## Bootstrap the overall system
 bootstrap: require-aws-profile require-cluster-mounted require-cluster-online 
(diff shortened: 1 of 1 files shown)


rootfs/usr/local/include/toolbox/Makefile
```diff
@@ -53,8 +53,8 @@ configure:
 .PHONY : bootstrap
 ## Bootstrap the overall system
 bootstrap: require-aws-profile require-cluster-mounted require-cluster-online 
-	@[ ! -f $(REMOTE_STATE)/.bootstrapped ] || (echo "Cluster already bootstraped"; exit 1)
-	@[ -z "$(DOCKER_USER)" ] || [ -z "$(DOCKER_PASSWORD)" ] || [ -z "$(DOCKER_EMAIL)" ] \
+	@[ ! -f $(REMOTE_STATE)/.bootstrapped ] || [echo "Cluster already bootstraped"; exit 1]
```
osterman: Why the brackets? This turns it into a conditional.

PR #50, opened 2017-07-02
Made kops respect default AWS_REGION

goruha:
## What
* Made kops respect default `AWS_REGION`

## Why
* When I specify default region for profile, I want kops to use it as default also

### DIFF
--- a/rootfs/usr/local/include/toolbox/kops/Makefile.defaults
+++ b/rootfs/usr/local/include/toolbox/kops/Makefile.defaults
@@ -1,6 +1,6 @@
 export KOPS_BIN ?= kops
 
-export KOPS_REGION ?= us-west-2
+export KOPS_REGION ?= $(AWS_REGION)
 
 export KOPS_DNS_ZONE ?= $(CLUSTER_DNS_ZONE)
 export KOPS_NAME ?= $(CLUSTER_NAME)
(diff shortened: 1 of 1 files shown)


rootfs/usr/local/include/toolbox/kops/Makefile.defaults
```diff
@@ -9,7 +9,7 @@ export KOPS_NAME ?= $(CLUSTER_NAME)
 export KOPS_STATE_STORE ?= s3://$(CLUSTER_STATE_BUCKET)
 
 export KOPS_CLOUD ?= aws
-export KOPS_ZONES ?= us-west-2a
+export KOPS_ZONES ?= $(AWS_REGION)a
```
osterman: This might not always work, since zones are not the same for all AWS accounts, but I don't have a better solution right now.

pureclouds: Hot sure how whole geodesic workflow was build but as I see in `Makefile.defaults` some variables set by result of external shell calls (just [this](https://github.com/cloudposse/geodesic/blob/master/rootfs/usr/local/include/toolbox/kops/Makefile.defaults#L22) example). 

So, why not use the same way to define first available AZ in region ? something like this:

```
export KOPS_ZONES ?= $(shell  aws ec2 describe-availability-zones  --filters Name=state,Values=available | jq -r '.AvailabilityZones[0].ZoneName')
```

>note: of course `aws cli` and `jq` should be installed into environment where `make` runs

PR #59, opened 2017-07-27
Use docker inheritance to define clusters

osterman:
## what
* Build one container per cluster
* Define essential ENVs in `Dockerfile`

## why
* previously it was too complicated to work with multiple cluster in the same shell
* sometimes clusters need to use different versions of geodesic
* allow multiple concurrent cluster sessions
* this model is a more scalable way to manage clusters and simpler to manage


pureclouds:
### Just small question about `geodesic`. bash script

[Here](https://github.com/cloudposse/geodesic/blob/bb87981f50ba2ce8847a4fbaf57a8d63e6668a54/rootfs/opt/geodesic#L61-L63) you set  options `--rm` and `--name`. But [later](https://github.com/cloudposse/geodesic/blob/bb87981f50ba2ce8847a4fbaf57a8d63e6668a54/rootfs/opt/geodesic#L106-L107) you add these options to array again without conditions. Of course it's not a bug, but looks not too ideal :)

osterman:
@pureclouds another great catch. Ya, that looks like it's redundant.

rootfs/opt/geodesic
```diff
@@ -53,6 +52,10 @@ elif [ "$1" == "uninstall" ]; then
 elif [ "$1" == "stop" ]; then
   echo "# Stopping ${DOCKER_NAME}..."
   exec docker kill ${DOCKER_NAME} >/dev/null 2>&1
+elif [ "$1" == "use" ]; then
+  export DOCKER_IMAGE=${DOCKER_IMAGE:-$2}
+  shift
+  shift
 fi
```
osterman: We need to validate that `DOCKER_IMAGE` is set.

README.md
```diff
@@ -57,10 +56,25 @@ Docker can be easily installed by following the instructions for your OS:
    ```
    curl -s https://geodesic.sh | bash
    ```
-2. Run `geodesic` to start the geodesic shell
-3. Run `setup-role` to configure your AWS credentials
-4. Run `assume-role $role` where $role is the one you configured in your AWS configuration.
-5. Run `ssh-add ~/.ssh/id_rsa` to add your ssh key to the SSH agent (must be same key used on Github)
+2. Create a `Dockerfile` that defines your environment
+   ```
```
osterman: Fix markdown

PR #60, opened 2017-07-29
Remove "hub" module references from toolbox

aknysh:
## What
* Remove "hub" module references from the main toolbox Makefile.
* Removed "cloud config use" command.

## Why
* Removes Geodesic dependencies on GitHub.
* Handling concurrent clusters in the same shell is not scalable and it was too complicated to work with multiple cluster in the same shell.
* Sometimes clusters need to use different Geodesic versions.
* Use Docker inheritance to define a cluster per shell instead.


aknysh:
Removed https://github.com/cloudposse/geodesic/blob/easy-clusters2/rootfs/usr/local/include/toolbox/hub/Makefile

.gitignore
```diff
@@ -1,2 +1,3 @@
 .build-harness
 build-harness/
+/.idea/
```
osterman: I think `.idea/` is sufficient, no?

aknysh: yes, sufficient

PR #61, opened 2017-08-03
Removed all references to CLOUD_CONFIG and CLOUD_CONFIG_SAMPLE vars

aknysh:
## What
* Removed all references to CLOUD_CONFIG and CLOUD_CONFIG_SAMPLE vars.

## Why
* These vars are not used anymore in the new geodesic version with Docker inheritance.

### DIFF
--- a/rootfs/etc/profile.d/defaults.sh
+++ b/rootfs/etc/profile.d/defaults.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # Setup some default envs
 export LESS=-Xr
 
@@ -75,7 +77,5 @@ export TF_BUCKET_PREFIX=geodesic/terraform
 #
 # Geodesic
 #
-export CLOUD_CONFIG=${REMOTE_STATE}/env
-export CLOUD_CONFIG_SAMPLE=${GEODESIC_PATH}/config/env.sample
 export SHELL_NAME=Geodesic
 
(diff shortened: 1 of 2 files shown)


osterman:
The `kops` and `kubernetes` modules depend on secrets that were typically stored in the `env` file. What should we do about that?

aknysh:
env.sample was already deleted in the previous commits.

The kops and kubernetes modules depend on secrets that were typically stored in the env file. What should we do about that?

In env.sample we had the following settings:
CLUSTER_DNS_ZONE=example.com                            # Parent zone for cluster (e.g. example.com)
CLUSTER_PREFIX=foobar                                   # Short name of cluster (e.g. foobar)
CLUSTER_STATE_BUCKET_REGION=us-east-1                   # Primary region of bucket
SSH_USERNAME=admin                                      # Username to use for connecting to cluster (e.g. github username)
DOCKER_REGISTRY_SERVER=https://index.docker.io/v1/      # Primary Docker Registry to use for pulling images
DOCKER_EMAIL=                                           # Email associated with Docker Registry account
DOCKER_USER=                                            # Username to authenticate with Docker Registry
DOCKER_PASSWORD=                                        # Password associated with Docker Username
GITHUB_ORG=                                             # Github Organization (or username) to host cluster repos

The first four are already in Dockerfile.
Would not it be better to have all the settings in one place? In Dockerfile?
So they are easy to maintain and setup/change for a specific cluster.

osterman:
Agreed. I think you have carried over the most important fields (top 4) to the `Dockerfile`.

PR #62, opened 2017-08-04
Fix ENV vars

aknysh:
## What
* Removed ```KOPS_MASTER_SIZE, KOPS_NODE_SIZE, KOPS_NODE_COUNT``` from toolbox/kops/Makefile.defaults.
* Fixed ```geodesic use --name``` in README.md

## Why
* The vars ```KOPS_MASTER_SIZE, KOPS_NODE_SIZE, KOPS_NODE_COUNT``` are cluster-specific (user-configurable) and should not be declared in the generic geodesic Docker image.
* ```geodesic use``` requires the name of the cluster specified via the ```name``` parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Docker can be easily installed by following the instructions for your OS:
    ```
 
 
-4. Run `geodesic use aws.example.com` to start the geodesic shell for that cluster
+4. Run `geodesic use --name aws.example.com` to start the geodesic shell for that cluster
 5. Run `cloud create` to run through configuration steps
 6. Run `cloud up` to provision the cluster
 
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Docker can be easily installed by following the instructions for your OS:
    ```
 
 
-4. Run `geodesic use --name aws.example.com` to start the geodesic shell for that cluster
+4. Run `geodesic use --name=aws.example.com` to start the geodesic shell for that cluster
 5. Run `cloud create` to run through configuration steps
 6. Run `cloud up` to provision the cluster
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -77,7 +77,7 @@ Docker can be easily installed by following the instructions for your OS:
    ```
 
 
-4. Run `geodesic use aws.example.com` to start the geodesic shell for that cluster
+4. Run `geodesic use --name aws.example.com` to start the geodesic shell for that cluster
```
osterman: needs to be `--name=aws.example.com`

rootfs/usr/local/include/toolbox/kops/Makefile.defaults
```diff
@@ -10,13 +9,8 @@ export KOPS_STATE_STORE ?= s3://$(CLUSTER_STATE_BUCKET)
 
 export KOPS_CLOUD ?= aws
 export KOPS_ZONES ?= $(AWS_REGION)a
-
-export KOPS_MASTER_SIZE ?= t2.medium
 export KOPS_MASTER_ZONES ?= $(KOPS_ZONES)
 
-export KOPS_NODE_SIZE ?= t2.medium
```
osterman: Why are we deleting the defaults here? If we delete them here, we should move them elsewhere. I'm okay with moving them to the `Dockerfile`

aknysh: The vars have been moved to Dockerfile, not from geodesic repo, but to the cluster-related repos.
We probably don't want anything cluster-specific (and user-configurable) to be in the generic geodesic image.

https://github.com/cloudposse/cluster-4.cloudposse.com/blob/master/Dockerfile
https://github.com/gadgetreview/aws.gadgetreview.com/blob/fix-env-vars/Dockerfile

osterman: @aknysh I have redacted your comment as it referred to private repositories. Keep in mind that this is a public repo.

osterman: They need to be documented/greppable. In the `Makefile`, it served this purpose. If we move them out of there, then we should put them somewhere else.

PR #65, opened 2017-08-10
Add Support for Native `USER`

s2504s:
## What
* Added script `user.sh` which creates the user with name `${USER}` if it does not exist in the container.

## Why
* The main goal is to create the user based on environment variable `${USER}` in the container.
* If the local user does not exist, this causes an error with use Ansible

### DIFF
--- a/rootfs/etc/profile.d/user.sh
+++ b/rootfs/etc/profile.d/user.sh
@@ -2,5 +2,7 @@
 
 id "${USER}" &>/dev/null
 if [[ "$?" -ne 0 ]]; then
-    adduser -D -H ${USER} &>/dev/null
+    if [[ -n "${USER_ID}" ]] && [[ -n "${GROUP_ID}" ]]; then
+        adduser -D -u ${USER_ID} -g ${GROUP_ID} -h ${HOME} ${USER} &>/dev/null
+    fi
 fi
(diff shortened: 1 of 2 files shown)


osterman:
Okay, so for some more context, the `geodesic` wrapper script passes in the `USER` env from the caller. 

https://github.com/cloudposse/geodesic/blob/master/rootfs/opt/geodesic#L63

@s2504s, you're saying that `ansible` uses this variable and panics if the user does not exist in the `/etc/passwd`?

rootfs/etc/profile.d/user.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+id "${USER}" &>/dev/null
+if [[ "$?" -ne 0 ]]; then
+    adduser -D -H ${USER} &>/dev/null
```
osterman: Can you use `HOME` as their home directory?

rootfs/opt/geodesic
```diff
@@ -60,7 +63,9 @@ function use() {
                        --env SSH_CLIENT
                        --env SSH_CONNECTION
                        --env SSH_TTY
```
osterman: Should we pass `HOME`?

s2504s: Let's use existing variable `${HOME}`

osterman:
- [ ] squash & merge
- [ ] tag `0.2.1`

PR #67, opened 2017-08-15
Bump terraform version to 0.10.1

comeanother:
## What

* Bump terraform version to 0.10.1

## Why

* Get ability to use latest terraform features

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ USER root
 WORKDIR /tmp
 
 # Install Terraform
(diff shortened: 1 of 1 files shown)


osterman:
image

comeanother:
@osterman should we do `rebase` instead of `merge` from master?

osterman:
Yes - you are right. Generally we should do a rebase against master. I just hit the update button which does a merge, but that is not ideal.

comeanother:
Tagged `0.3.0`

PR #69, opened 2017-08-21
Ability for environment configuration in Development mode

s2504s:
## What
* Added a Development mode flag
* Added a shell script that allows to configure the environment in Development mode

## Why
* The development mode can be configured with different environments and additional functionality

### DIFF
--- a/rootfs/etc/profile.d/dev-mode.sh
+++ b/rootfs/etc/profile.d/dev-mode.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+if [ "${DEV}" == "true" ]; then
+    echo "Installation of Ansible dependencies for the Development mode..."
+    make -C /conf/ansible deps
+fi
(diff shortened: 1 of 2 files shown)


rootfs/etc/profile.d/dev-mode.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
osterman: We should rename this to `ansible.sh`

rootfs/etc/profile.d/dev-mode.sh
```diff
@@ -0,0 +1,6 @@
...
+
+if [ "${DEV}" == "true" ]; then
+    echo "Installation of Ansible dependencies for the Development mode..."
```
osterman: This should only happen if `/conf/ansible/Makefile` exists. E.g. `geodesic` is used for all kinds of projects, some of which do not use `ansible`

osterman:
Actually, I prefer we drop the ansible extensions from this PR. It's okay if we keep the `DEV` changes, but I think the `ansible` stuff is too specific.

osterman:
This is not relevant to `geodesic` and is too proprietary.

PR #74, opened 2017-09-25
Add default ansible.cfg

SweetOps:
## What

* Add default ansible.cfg

## Why

* ansible.cfg can be absent in playbook directory

### DIFF
--- a/rootfs/etc/ansible/ansible.cfg
+++ b/rootfs/etc/ansible/ansible.cfg
@@ -1,0 +1,6 @@
+[defaults]
+roles_path = ./roles
+inventory = ./inventory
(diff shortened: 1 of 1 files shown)


comeanother:
`ansible.cfg` calculates relative paths starting from directory where `ansible.cfg`  is being placed.

We may never have `roles` and `playbooks` folders within `rootfs/etc/ansible/`.

Usually, work directory tree is
```
├── conf
│   ├── ansible
│   │   ├── ansible.cfg
│   │   ├── inventory
│   │   ├── playbooks
│   │   ├── roles
│   │   └── scripts
│   └── tf
```

PR #80, opened 2017-11-07
Swap out s3fs with goofys

aknysh:
## what
* Replaced `s3fs` with `goofys`

## why
* `goofys` is better alternative to `s3fs` written in `Go` that natively supports assumed roles

## references
* https://github.com/kahing/goofys


aknysh:
i'm closing it, it has too many changes, will decompose into smaller PRs

PR #81, opened 2017-11-09
Add gomplate

osterman:
## what 
* Add `gomplate` templating engine

## why
* Support parameterized templates with a more powerful templating language than provided by [envsubst]( https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html)


Dockerfile
```diff
@@ -76,7 +76,7 @@ RUN apk --update add fuse libxml2 mailcap && \
     apk del .build-deps
 
 # Install Google Cloud SDK
-ENV GCLOUD_SDK_VERSION=147.0.0
+ENV GCLOUD_SDK_VERSION=179.0.0
```
osterman: Can we upgrade packages as a separate PR

templates/kops/default.yaml
```diff
@@ -0,0 +1,126 @@
+apiVersion: kops/v1alpha2
```
osterman: Let's add `kops` manifest as a separate PR. I need `gomplate` to do what I'm working on.

PR #101, opened 2017-12-13
Fix HELM HOME

goruha:
## What
* Fix `HELM_HOME` env var

## Why
* On docker build helm plugins installed into directory that are not used later

### DIFF
--- a/rootfs/etc/profile.d/defaults.sh
+++ b/rootfs/etc/profile.d/defaults.sh
@@ -31,7 +31,6 @@ fi
 #
 # Helm
 #
-export HELM_HOME=${REMOTE_STATE}/helm/
 export HELM_VALUES_PATH=${REMOTE_STATE}/helm/values
 
 #


rootfs/etc/profile.d/defaults.sh
```diff
@@ -31,7 +31,6 @@ fi
 #
 # Helm
 #
-export HELM_HOME=${REMOTE_STATE}/helm/
```
osterman: The problem is we install helm plugins during build but have a `HELM_HOME` that is expected to be mounted.

I think it should be instead:
 ```
ENV HELM_HOME /var/lib/helm
```

We want a consistent location for these. By not setting it, it will use `/root/.helm`.

goruha: hm... 
We define `HELM_HOME` on docker build 
https://github.com/cloudposse/geodesic/blob/fix-helm-home/Dockerfile#L75
When `/etc/profile.d/defaults.sh` runs `HELM_HOME` is already defined and point to `/var/lib/helm`

I do not see any reason why we should override it in `/etc/profile.d/defaults.sh`

goruha: Why we need `HELM_HOME` to be mounted?

PR #103, opened 2018-01-08
README: Modify docker command to inline variable

rtizzy:
## what
* Inline the `CLUSTER_NAME` variable

## why
* Users utilizing sudo to run Docker need to inline the variable as this is not inherited by sudo
* Running docker with sudo is generally considered best practice since the "docker" group sometimes suggested to run docker as a non-root user is equivalent to root privileges. 

https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface


osterman:
Closing for now

Issue #112, opened 2018-04-05
support for zsh


rohit-zabbed:
## what
* Do not require `bash4`
* Would be great if we can support `zsh` as well.

## why
* The default script requires bash4 which enforce to have separate bash installation (e.g. via `brew`)

osterman:
I agree. I want to drop this requirement. =)

Might be simplifying/refactoring the wrapper script to accommodate that. Thanks for speaking up!

osterman:
@rohit-zabbed we've restored compatibility with bash 3, so it's no longer necessary to install `bash4`

osterman:
@rohit-zabbed can you elaborate on the zsh support you want?

osterman:
Bash for the foreseeable future will remain the primary language for scripting integrations between all the various tools.

PR #119, opened 2018-04-08
Backwards Compatibility for Bash v3

osterman:
## what
* Remove bash4 syntax (e.g. `declare -g` and `${k^^}`)
* Fix mounting of `$HOME` for linux systems (regression)

## why
* Requiring `bash4` increases barrier of adoption without adding significant benefits. Minor drawback is we require `tr` instead.

## references
* https://github.com/cloudposse/geodesic/issues/112


osterman:
@goruha: @atedja confirmed this works for him.

PR #123, opened 2018-04-18
Sample for using ONBUILD

rverma-nikiai:
Was age to reduce size from 1.01GB to 750 Mb using
#docker build --rm --build-arg KUBERNETES_VERSION=1.9.6 --build-arg KOPS_VERSION=1.9.0 --build-arg AWS_VAULT_VERSION="" -t nikiai/tools .

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,24 @@
-FROM alpine:3.7
+FROM alpine:3.6
 
-ENV BANNER "geodesic"
+ONBUILD ARG BANNER="geodesic"
+
+# Install all packages as root
+USER root
+
+RUN echo -e "http://nl.alpinelinux.org/alpine/v3.6/main\nhttp://nl.alpinelinux.org/alpine/v3.6/community" > /etc/apk/repositories
+# Install common packages
+RUN apk add --no-cache --update unzip curl tar \
+          python make jq figlet \
+          openssl openssh-client sshpass iputils drill \
+          gcc libffi-dev python-dev musl-dev openssl-dev py-pip py-virtualenv \
+          git coreutils less groff bash-completion \
+          fuse syslog-ng libc6-compat && \
+          rm -rf /tmp/* /var/cache/apk/* && \
+          mkdir -p /etc/bash_completion.d/ /etc/profile.d/ \
+    && mkdir -p /conf \
+    && touch /conf/.gitconfig
+
+RUN echo "net.ipv6.conf.all.disable_ipv6=0" > /etc/sysctl.d/00-ipv6.conf
 
 # Where to store state
 ENV CACHE_PATH=/localhost/.geodesic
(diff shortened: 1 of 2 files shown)


osterman:
This is a really great concept. So basically a user can decide which tools to bundle rather than shipping 100% of things in the base image.

Dockerfile
```diff
@@ -1,46 +1,50 @@
-FROM alpine:3.7
+FROM alpine:3.6
```
osterman: Why the downgrade?

rverma-nikiai: Faced some issue with bash apk's

rverma-nikiai:
In my last update, as pre suggestion from docker community, we can leverage global args. This way we don't even need to pass any build args.

Dockerfile
osterman: hrmm shouldn't need to downgrade. Can you share the error you had? I bet I've seen it before.

Dockerfile
```diff
@@ -1,46 +1,50 @@
...
+
+ENV GEODESIC_PATH=/usr/local/include/toolbox
+ENV MOTD_URL=http://geodesic.sh/motd
+ENV HOME=/conf
+ENV KOPS_CLUSTER_NAME=example.foo.bar
+ENV SECRETS_PATH=${HOME}
+
 # Disable vim from reating a swapfile (incompatible with goofys)
-RUN echo 'set noswapfile' >> /etc/vim/vimrc
+#RUN echo 'set noswapfile' >> /etc/vim/vimrc
```
osterman: the reason for this is if you `vim` a file on `goofys` filesystem, the `.swp` file generated by vim will cause problems.

example/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+#ARG ANSIBLE_VERSION
+#ARG GCLOUD_SDK_VERSION
+#ARG AWSEBCLI_VERSION
+#ARG AWSCLI_VERSION
+#--------------------------------------
+
+FROM nikiai/geodesic
+
+ARG AWLESS_VERSION=0.1.10
+RUN curl --fail -SL -O https://github.com/wallix/awless/releases/download/${AWLESS_VERSION}/awless-linux-amd64.tar.gz \
```
osterman: Looks neat

example/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    && tar -xzf awless-linux-amd64.tar.gz \
+    && rm awless-linux-amd64.tar.gz \
+    && mv awless /usr/local/bin
+RUN apk add --no-cache py2-lxml
+RUN pip install aws-google-auth
```
osterman: Suggest version pinning =)

example/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG BANNER="niki geodesic"
+ARG KUBERNETES_VERSION=1.9.6
+ARG KOPS_VERSION=1.9.0
+
+# other args than can be overridien are
```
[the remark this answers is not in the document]
osterman: Also, maybe add something like: 
> NOTE: setting an empty version string will disable the installation of the app

Dockerfile
```diff
@@ -196,16 +206,19 @@ RUN curl --fail -sSL -o /usr/local/bin/goofys https://github.com/kahing/goofys/r
 #
 # Install Google Cloud SDK
 #
-ENV GCLOUD_SDK_VERSION=179.0.0
-RUN curl --fail -sSL -O https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${GCLOUD_SDK_VERSION}-linux-x86_64.tar.gz && \
+ONBUILD ARG GCLOUD_SDK_VERSION
+#179.0.0
```
osterman: I'm torn. I think it might be best to let the caller *disable* a feature by setting a nil arg, rather than leaving it disabled by default.

rverma-nikiai: Works that way also, or you can cherry pick the behavior. Some needs enabling e.g Ansible (else by default it will be skipped) on the other hand (some might need disabling).
This would make script bit complicated though as you need to keen relationship between 2.
Let's say a user can use dep A, but if he wants dep B (A can't be excluded ). This kind of scenario becomes tricky[though remember YAGNI]. Another thing which I want to do but couldn't do is, if we provide the only arg as KOPS_VERSION[use some pinned persion], if we use KOPS_VERSION=1.9, use this version. Something like requirements.txt kind of behavior.

Dockerfile
rverma-nikiai:
```ERROR: ncurses-terminfo-6.0_p20171125-r0: Protocol error
(6/77) Installing readline (7.0.003-r0)
(7/77) Installing bash (4.4.19-r1)
ERROR: bash-4.4.19-r1: Protocol error
(8/77) Installing bash-completion (2.7-r3)
...
(14/77) Installing libcurl (7.59.0-r0)
ERROR: libcurl-7.59.0-r0: Protocol error
(15/77) Installing curl (7.59.0-r0)
...
(17/77) Installing drill (1.6.17-r6)
ERROR: ldns-1.6.17-r6: Protocol error
```

rootfs/etc/profile.d/aws-google-auth.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+if [ -n "${AWS_GOOGLE_AUTH}" ]; then
```
osterman: Yea, this is good. I'll do something like this for `AWS_VAULT`

goruha:
@rverma-nikiai @osterman  - I suggest to decline this pattern

osterman:
@rverma-nikiai curious about your thoughts on this: https://github.com/cloudposse/geodesic/pull/164

rverma-nikiai:
I actually loved the previous syntax, it gave a feel of package manager and
reduce layers in build.

This is reversed.

osterman:
>  it gave a feel of package manage

Check it out now. Best of both worlds. The `cloudposse/packages` repo is where we maintain binary releases of everything. It's almost like a linux distro for arbitrary apps. Everything/anything can be added to this image. We'll accept PRs from you. Also, we have added `awless`, `kubens`, and `kubectx` per your recommendations. We reuse these packages in our `build-harness` and local `dev` harness, to name a few. Different use-cases neeed different packages.

Then using the `PACKAGES` build arg, the packages that get installed can be customized. 

> reduce layers in build

Consider `docker build --squash` =)

PR #143, opened 2018-06-10
Only use --privileged on OSX

osterman:
## what
* Only run `--privileged` mode on OSX

## why
* Docker on Linux (native) won't suffer from time drift so we don't need this hack


osterman:
We didn't merge this b/c it breaks aws-vault server mode. We should probably disable server mode automatically if not running in privileged mode (and warn about it).

osterman:
Closing this for now because I don't know how to best address this issue.

PR #147, opened 2018-06-14
geodesic_prompt(): add PROMPT_STYLE handler

### LINKED ISSUE #150: Add Support for Custom Prompts
## what
* Add support for `PROMPT_STYLE` environment variable that alters the behavior of the shell prompt

## why
* The current utf8 heavy prompt breaks some terminals
* Users have mentioned they'd prefer a plain prompt

tamsky:
## what
* Add support for a "plain" prompt 

## why
* UTF8 glyphs break some peoples terminals

Fixes #150

### DIFF
--- a/rootfs/etc/profile.d/prompt.sh
+++ b/rootfs/etc/profile.d/prompt.sh
@@ -30,27 +30,26 @@ PROMPT_HOOKS+=("geodesic_prompt")
 function geodesic_prompt() {
        
   case $PROMPT_STYLE in
-      plain)
-          # 8859-1 codepoints:
-          WHITE_HEAVY_CHECK_MARK=$(tput bold)$(tput setab 2)$'X'$(tput sgr0)' '
-          BLACK_RIGHTWARDS_ARROWHEAD=$'=> '
-          TWO_JOINED_SQUARES=$'¤ '  # perhaps § instead?
-          CROSS_MARK=$'× '
-          ;;
-      *)
-          # unicode
-          WHITE_HEAVY_CHECK_MARK=$'\u2714 '     # '✔'
-          BLACK_RIGHTWARDS_ARROWHEAD=$'\u27A4 ' # '➤', suggest '▶' may be present in more fonts
-          TWO_JOINED_SQUARES=$'\u29C9 '         # '⧉'
-          CROSS_MARK=$'\u274C '                 # '❌'
-          ;;
+    plain)
+      # 8859-1 codepoints:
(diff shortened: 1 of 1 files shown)


rootfs/etc/profile.d/prompt.sh
```diff
@@ -28,12 +28,25 @@ function terraform_prompt() {
...
+          CROSS_MARK=$'\u274C '                 # '❌'
+          ;;
+  esac
 
-  WHITE_HEAVY_CHECK_MARK=$'\u2705 '
-  BLACK_RIGHTWARDS_ARROWHEAD=$'\u27A4 '
-  TWO_JOINED_SQUARES=$'\u29C9 '
-  CROSS_MARK=$'\u274C '
-
+  # why do we 'export' STATUS?
```
osterman: This might be legacy. Don't think the `export` is needed.

osterman: if it works without it, drop it.

rootfs/etc/profile.d/prompt.sh
```diff
@@ -28,12 +28,25 @@ function terraform_prompt() {
 # Define our own prompt
 PROMPT_HOOKS+=("geodesic_prompt")
 function geodesic_prompt() {
+       
+  case $PROMPT_STYLE in
+      plain)
+          # 8859-1 codepoints:
+          WHITE_HEAVY_CHECK_MARK=$(tput bold)$(tput setab 2)$'X'$(tput sgr0)' '
+          BLACK_RIGHTWARDS_ARROWHEAD=$'=> '
+          TWO_JOINED_SQUARES=$'¤ '  # perhaps § instead?
```
osterman: Yea, I like `§`

osterman: perhaps we should generalize the names? e.g. it's no longer "Two joined names"

rootfs/etc/profile.d/prompt.sh
```diff
@@ -28,12 +28,25 @@ function terraform_prompt() {
 # Define our own prompt
 PROMPT_HOOKS+=("geodesic_prompt")
 function geodesic_prompt() {
+       
+  case $PROMPT_STYLE in
```
osterman: Perfect! thanks for making it a `case` statement.

rootfs/etc/profile.d/prompt.sh
tamsky: +1

I'll try my hand at some more generic sounding names?

rootfs/etc/profile.d/prompt.sh
tamsky: dropped.

tamsky:
example rendering:
image

PR #158, opened 2018-06-20
add DataDog useHostPort and updateStrategy values to master helmfile

daveyu:
## What
* Bump DataDog chart version
* Add `daemonset.useHostPort` value
* Add `daemonset.updateStrategy` value

## Why
* The latest chart version supports newer versions of the DataDog agent
* `daemonset.useHostPort` set to `true` allows trace agents from pods on the same node to report using the `hostIP`
* `daemonset.updateStrategy` can be set to `RollingUpdate` to relaunch DataDog agent pods when the chart is installed or updated


osterman:
Thanks @daveyu !

Issue #155, opened 2018-06-19
Upgrade Helm


osterman:
## what
* Use `helm` from `packages

## references
- <https://github.com/cloudposse/packages/issues/12>

PR #159, opened 2018-06-20
Helm - bump version

alebabai:
## what
* use helm 2.9.1

## why
* #155


osterman:
Move to https://github.com/cloudposse/geodesic/blob/master/Dockerfile#L44

Dockerfile
```diff
@@ -131,15 +131,9 @@ RUN curl --fail -sSL -o /usr/local/bin/sops https://github.com/mozilla/sops/rele
...
-ENV HELM_VERSION 2.8.2
 ENV HELM_HOME /var/lib/helm
 ENV HELM_VALUES_PATH=${SECRETS_PATH}/helm/values
-RUN curl --fail -sSL -O http://storage.googleapis.com/kubernetes-helm/helm-v${HELM_VERSION}-linux-amd64.tar.gz \
-    && tar -zxf helm-v${HELM_VERSION}-linux-amd64.tar.gz \
-    && mv linux-amd64/helm /usr/local/bin/helm \
-    && rm -rf linux-amd64 \
-    && chmod +x /usr/local/bin/helm \
-    && helm completion bash > /etc/bash_completion.d/helm.sh \
+RUN helm completion bash > /etc/bash_completion.d/helm.sh \
```
osterman: Should we maybe check if helm installed then run all these steps?

alebabai: We rely on packages layer, that it installed helm.
But it's possible to do inline if/else with check and failure (to fail the build)

osterman: so, `PACKAGES` is a build arg. That means it can be overridden by the caller. One "aspiration" I have for `geodesic` is to allow it to be more customizable by others. Since `helm` is only a requirement for kubernetes and not for ECS or using terraform, I want it to be possible to build mutltiple version of geodesic that are possibly lighter weight.

osterman: can't we just do something like:
```
[ -z `which helm` ] || helm version
```

PR #167, opened 2018-07-02
Add absolute path to s3fs in script

sarkis:
## what
Add the absolute path to the s3fs script for mounting s3 bucket via goofys.

## why
Fix a bug where pathing issue causes any `mount -a` invocation to throw an error:

```
/bin/sh: s3fs: not found
```


sarkis:
> So the mount command does not respect PATH?
@osterman I need to dig into this further, not exactly sure where in the flow `/usr/local/bin` is not in the PATH.

rootfs/usr/local/bin/s3
```diff
@@ -5,7 +5,7 @@ BUCKET=$2
 BUCKET_PREFIX=$3
 MOUNT_POINT=$4
 FSTAB="/etc/fstab"
-FSTYPE="s3fs"
+FSTYPE="/usr/local/bin/s3fs"
```
osterman: The other fix might be to move `s3fs` to `rootfs/usr/bin`

sarkis: Not against this, it seems like a cleaner solution to just have it in PATH somewhere for future invocations - so s3fs script works as intended every time. I can take a look at this by EOD.

osterman: crap - looks like we don't. I think we used to. Must have been lost. Let's add it to the `Dockerfile`.

sarkis: @osterman I tried this https://github.com/cloudposse/geodesic/pull/167/commits/4ae1e96665f6930750fc5f81c9af1ca4d4f6431d

Although, it does look like `PATH` at some point does have /usr/local/bin appended to it. The issue seems to be mount -a is not respecting the `PATH` env var. So far the only fix I've been  able to successfully apply has been absolute path in the fstab entry.

Dockerfile
```diff
@@ -48,7 +48,7 @@ RUN echo 'set noswapfile' >> /etc/vim/vimrc
 WORKDIR /tmp
 
 COPY --from=packages /dist/ /usr/local/bin/
-
+ENV PATH="/usr/local/bin:${PATH}"
```
osterman: Did this fix it?

rootfs/usr/local/bin/s3
osterman: If not, then let's move `s3fs` to `/usr/bin`

sarkis:
@osterman this seems like the best solution - thanks for the guidance... tested by building out the container and `mount -a`:

```
2018-07-10T15:23:11+00:00 ccbdfc8649d5 syslog-ng[154]: syslog-ng starting up; version='3.9.1'
2018-07-10T15:23:22+00:00 ccbdfc8649d5 /usr/local/bin/goofys[173]: main.INFO File system has been successfully mounted.
```

PR #168, opened 2018-07-02
Fixing outdated syntax for Prometheus PVCs

MoonMoon1919:
## What it is 
Previously this used an old spec that caused newer installations of the this chart to fail. `storageSpec` has since been updated to just `storage`

See:
https://github.com/coreos/prometheus-operator/issues/860#issuecomment-370999129

## How I Tested

Modified this file and deployed the kube-prometheus chart to a cluster.


osterman:
@goruha @aknysh heads up

PR #170, opened 2018-07-03
kiam agents should use iptables to intercept traffic

osterman:
## what
* Force iptables DNAT rule to reroute `169.254.169.254` to the `kiam` agent metadata proxy

## why
* Kiam stopped working; not entirely sure why
* Seems to have happened after upgrading to k8s 1.9.1

> This is the process that would typically be deployed as a DaemonSet to ensure that Pods have no access to the AWS Metadata API. Instead, the agent runs an HTTP proxy which intercepts credentials requests and passes on anything else. An DNAT iptables rule is required to intercept the traffic. The agent is capable of adding and removing the required rule for you through use of the --iptables flag. This is the name of the interface where pod traffic originates and it is different for the various CNI implementations. The flag also supports the ! prefix for inverted matches should you need to match all but one interface.

## references
* https://github.com/uswitch/kiam#agent

### DIFF
--- a/rootfs/conf/kops/helmfile.yaml
+++ b/rootfs/conf/kops/helmfile.yaml
@@ -70,7 +70,7 @@ releases:
 
     # show debug messages
     - name: "env.open.DEBUG"
-      value: '{{ env "CHARTMUSEUM_DEBUG" | default "true" }}'
+      value: '{{ env "CHARTMUSEUM_DEBUG" | default "false" }}'
 
     # allow chart versions to be re-uploaded
     - name: "env.open.ALLOW_OVERWRITE"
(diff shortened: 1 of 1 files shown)


rootfs/conf/kops/helmfile.yaml
```diff
@@ -68,6 +68,10 @@ releases:
     - name: "env.open.DISABLE_API"
       value: "true"
 
+    # show debug messages
+    - name: "env.open.DEBUG"
+      value: '{{ env "CHARTMUSEUM_DEBUG" | default "true" }}'
```
sarkis: Should this be defaulting to true here?

osterman: Yea, I suppose we should make it false. Thanks!

Issue #157, opened 2018-06-20
Decompose `helmfile.yaml` to `helmfile.d`


osterman:
## what
* https://github.com/cloudposse/geodesic/blob/master/rootfs/conf/kops/helmfile.yaml

## why
* More maintainable

## related
- <https://github.com/roboll/helmfile/pull/160>
- <https://github.com/roboll/helmfile/issues/137>
- <https://github.com/roboll/helmfile/issues/151>

- [ ] Update documentation for for `helmfile.d`

osterman:
Depends on a new release of https://github.com/roboll/helmfile/releases

osterman:
https://github.com/cloudposse/packages/pull/19

osterman:
We actually already released it, but had to [rollback due to issues with `helmfile`](https://github.com/cloudposse/geodesic/releases/tag/0.11.6).

For additional context, see:
- <https://github.com/roboll/helmfile/pull/190>
- <https://github.com/roboll/helmfile/pull/188>
- <https://github.com/roboll/helmfile/issues/187>
- <https://github.com/roboll/helmfile/issues/189>

Just today they merged the fixes upstream in `helmfile`. Now we are waiting for them to cut a new release. As soon as they do, we'll bring this back.

- <https://github.com/roboll/helmfile/releases>

osterman:
@maver1ck roboll/helmfile released with the fixes, so we're now unblocked

osterman:
https://github.com/cloudposse/helmfiles/pull/3

osterman:
@maver1ck have a look here: https://github.com/cloudposse/helmfiles

osterman:
@maver1ck new helmfile contributions welcome! we want to grow this library.

PR #171, opened 2018-07-04
Use helmfile.d feature

alebabai:
## what
Use `helmfile.d` feature to decompose `helmfile.yaml`

## why
#157


alebabai:
It seems to me that helmfile don't see helmfile.d folder by default
https://github.com/roboll/helmfile/pull/160#issuecomment-402453737

alebabai:
@osterman also i decomposed charts by _component_ label. What do you think about this idea?

osterman:
Let's stick each chart in it's own helmfile because it allows people to easily overwrite a single file

osterman:
it's a good idea to use the `01.` and `02.` prefix, but it's also good to separate them by some gap. this allows changing the order without needing to renumber all files.

alebabai:
> Let's stick each chart in it's own helmfile because it allows people to easily overwrite a single file
ok, got it.

> it's a good idea to use the 01. and 02. prefix, but it's also good to separate them by some gap. this allows changing the order without needing to renumber all files.
Didn't quite catch what you meant.
Do you prefer to use something like this `01_portal.yaml` or just add number if it realy needed?

osterman:
Yes, I think your screenshot captures it. We want some gaps so it's less painful to move things around.

PR #173, opened 2018-07-05
Use cloudposse fluentd chart

goruha:
## What
* Switch `fluentd` chart to `fluentd-kubernetes`

## Why
* Use generic chart
* It's going to be a while before [official chart PR](https://github.com/kubernetes/charts/pull/5357) gets approved

### DIFF
--- a/rootfs/conf/kops/helmfile.d/0510.fluentd.yaml
+++ b/rootfs/conf/kops/helmfile.d/0510.fluentd.yaml
@@ -7,9 +7,9 @@ helmDefaults:
     - "--reset-values"
 
 repositories:
-# Incubator repo of official helm charts
-- name: "incubator"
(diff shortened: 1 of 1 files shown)


rootfs/conf/kops/helmfile.d/0510.fluentd.yaml
```diff
@@ -28,11 +28,11 @@ releases:
 - name: "logs"
   namespace: "kube-system"
   labels:
-    chart: "fluentd"
+    chart: "fluentd-kubernetes"
     component: "datadog"
     namespace: "kube-system"
     default: "false"
-  chart: "incubator/fluentd"
+  chart: "cloudposse-incubator/fluentd-kubernetes"
```
osterman: `cloudposse-incubator` not defined in `repositories`.

PR #181, opened 2018-07-18
portal ingress.tls.enabled from env

daveyu:
## What
Optionally read the value for the portal's `ingress.tls.enabled` from an environment variable

## Why
Allow the portal chart to be installed in scenarios where TLS is not used.


daveyu:
@osterman OK if I cut a release for this?

osterman:
👍

PR #184, opened 2018-07-19
use CHARTMUSEUM_API_HOSTNAME

daveyu:
## What
Use env var CHARTMUSEUM_API_HOSTNAME for the chartmuseum API release instead of `api.{{ env "CHARTMUSEUM_HOSTNAME" }}`

## Why
Allow other host naming conventions.

In my case, I want to use `api-charts.us-west-2.staging.example.com` as the hostname so that a `*.us-west-2.staging.example.com` certificate applies.

## upgrading

1. Define `CHARTMUSEUM_API_HOSTNAME` with `chamber` (e.g. `api.charts.us-west-2.staging.cloudposse.org`)


osterman:
@daveyu I've updated your description with upgrade instructions. Look good?

Issue #199, opened 2018-07-30
Fix `$HOME` Mounting for WSL


osterman:
## what
* Mounting `$HOME` to `/localhost` is not working on WSL (Windows Shell for Linux)

## why
* Explain why this is a problem and what is the expected behavior.
* Explain why this feature request or enhancement is beneficial.

## examples
This works:
```
docker run -it --rm -v /C/Users/sebas_000/AppData/Local/lxss/home/martaver:/test alpine sh
```

This does not:
```
docker run -it --rm -v /home/martaver:/test alpine sh
```

Maybe we should do something like in the wrapper script if we detect WSL:
```
mount --bind /mnt/c /c
```
or 
```
mount --bind /C/Users/sebas_000/AppData/Local/lxss/home/martaver /home/martaver
```

## references
- <https://github.com/Microsoft/WSL/issues/1854>
- <https://superuser.com/questions/1344407/docker-on-wsl-wont-bind-mount-home>
- <https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly>

Martaver:
There's a current explicit workaround that treats WSL as a special case. The make install script needs to be updated to ensure the vendor script uses the following at line 92:

```
echo "# Mounting ${HOME} into container"      
      if  [[ $(cat /proc/version) =~ Microsoft ]] ;
      then        
        localHome=/$(cmd.exe /c echo %LOCALAPPDATA% | tr -d '\r' | sed -e 's/\\/\//g' -e 's/://')/lxss$HOME
        echo Detected Windows Subsystem for Linux, mounting $localHome instead of $HOME
      else
        localHome=$HOME
      fi
      DOCKER_ARGS+=(--volume=${localHome}:/localhost)
```

Martaver:
This script detects WSL, and if it is, figures out the local (windows) directory path to $HOME. 

This works because Docker for Windows can see /c as a volume. Using Docker for Windows on hyper-v, hyper-v does not mount any volume as /home/username as in $HOME, and so using the plain $HOME var from WSL fails.

vadim-hleif:
@Martaver Hi,
- Why do you use lxss directory? 
- Did you install wsl from market? 
- Which version of wsl do you use?
### references
- https://stackoverflow.com/questions/39886862/the-lxss-folder-in-windows-for-bash-on-ubuntu-on-windows-is-missing/47193112#47193112

Issue #196, opened 2018-07-29
Add Support for Windows (WSL/Ubuntu Linux)


osterman:
## what
* Add support for running Geodesic shells on [Windows using WSL](https://docs.microsoft.com/en-us/windows/wsl/about)

## why
* We have some early adopters running windows
* Minimal effort seems required to get this working on windows (seems very close!)

## references
*  https://github.com/cloudposse/root.cloudposse.co/issues/6
* https://github.com/cloudposse/geodesic/issues/199

vadim-hleif:
Implemented

aknysh:
addressed in https://github.com/cloudposse/geodesic/pull/202

PR #202, opened 2018-07-31
Add Support for Windows Subsystem for Linux (WSL)

vadim-hleif:
## what
* `$HOME` path now is built dynamically  (search wsl dir on windows, getting windows and linux users names)
## why
* By default `$HOME` refers on `/usr/local/bin`, but in case with `WSL`, docker installed on windows system and this path not correct

## references
- <https://stackoverflow.com/a/47193112>
- <https://github.com/cloudposse/geodesic/issues/199>
- <https://github.com/cloudposse/root.cloudposse.co/issues/8>
- <https://github.com/cloudposse/geodesic/issues/196> <https://www.reddit.com/r/bashonubuntuonwindows/comments/8dhhrr/is_it_possible_to_get_the_windows_username_from/>

### DIFF
--- a/rootfs/templates/wrapper
+++ b/rootfs/templates/wrapper
@@ -89,6 +89,11 @@ function use() {
     if [ "${HOME}" == "/localhost" ]; then
       echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
     else
+      if [[ $(uname -r) =~ Microsoft$ ]]; then
+        WSL_USER_DIR=${WSL_USER_DIR:-$(ls -d /mnt/c/Users/* | head -1)} 
+        WSL_USER_NAME=${WSL_USER_NAME:-user} 
+        HOME="${WSL_USER_DIR}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc/LocalState/rootfs/home/${WSL_USER_NAME}"
+      fi
       echo "# Mounting ${HOME} into container"
       DOCKER_ARGS+=(--volume=${HOME}:/localhost)
     fi
--- a/rootfs/templates/wrapper
+++ b/rootfs/templates/wrapper
@@ -84,18 +84,29 @@ function use() {
     # Implement DNS fix related to https://github.com/docker/docker/issues/24344
     DOCKER_ARGS+=("--dns=${DOCKER_DNS}")
   fi
-
+     
   if [ -n "${HOME}" ]; then
     if [ "${HOME}" == "/localhost" ]; then
       echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
     else
-      if [[ $(uname -r) =~ Microsoft$ ]]; then
-        windows_username=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | sed -e 's/\r//g')
-        wsl_app_dir=$(find /mnt/c/Users/${windows_username}/AppData/Local/Packages/ -maxdepth 1 -name "CanonicalGroupLimited\.*Ubuntu*onWindows*" | sed 's:.*/::')
-        HOME="C:/Users/${windows_username}/AppData/Local/Packages/${wsl_app_dir}/LocalState/rootfs/home/${USER}"
+      if  [[ $(cat /proc/version) =~ Microsoft ]] ; 
+      then
+        windowsUserName=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | tr -d '\r')
+
+        if [ -d "/mnt/c/Users/${windowsUserName}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc" ]; 
+        then 
+          localHome=C:/Users/${windowsUserName}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc/LocalState/rootfs${HOME}
+        elif [ -d "/mnt/c/Users/${windowsUserName}/AppData/Local/lxss/" ]
+        then 
+          localHome=C:/Users/${windowsUserName}/AppData/Local/lxss${HOME}
+        fi
+
+        echo Detected Windows Subsystem for Linux, mounting $localHome instead of $HOME
+      else
+        localHome=$HOME
       fi
-      echo "# Mounting ${HOME} into container"
-      DOCKER_ARGS+=(--volume=${HOME}:/localhost)
+      echo "# Mounting ${localHome} into container"
+      DOCKER_ARGS+=(--volume=${localHome}:/localhost)
     fi
   fi
 
--- a/rootfs/templates/wrapper
+++ b/rootfs/templates/wrapper
@@ -90,9 +90,8 @@ function use() {
       echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
     else
       if [[ $(uname -r) =~ Microsoft$ ]]; then
-        WSL_USER_DIR=${WSL_USER_DIR:-$(ls -d /mnt/c/Users/* | head -1)} 
-        WSL_USER_NAME=${WSL_USER_NAME:-user} 
-        HOME="${WSL_USER_DIR}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc/LocalState/rootfs/home/${WSL_USER_NAME}"
+        windows_username=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | sed -e 's/\r//g')
+        HOME="C:/Users/${windows_username}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc/LocalState/rootfs/home/${USER}"
       fi
       echo "# Mounting ${HOME} into container"
       DOCKER_ARGS+=(--volume=${HOME}:/localhost)
--- a/rootfs/templates/wrapper
+++ b/rootfs/templates/wrapper
@@ -87,7 +87,7 @@ function use() {
 
   if [ -n "${LOCAL_HOME}" ]; then
     local_home=${LOCAL_HOME}
-  elif [[ $(cat /proc/version) =~ Microsoft ]]; then
+  elif [[ $(uname -r) =~ Microsoft$ ]]; then
     windows_user_name=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | tr -d '\r')
     user_local_app_data=$(cmd.exe /c echo %LOCALAPPDATA% | tr -d '\r' | sed -e 's/\\/\//g')
 
--- a/rootfs/templates/wrapper
+++ b/rootfs/templates/wrapper
@@ -84,51 +84,38 @@ function use() {
     # Implement DNS fix related to https://github.com/docker/docker/issues/24344
     DOCKER_ARGS+=("--dns=${DOCKER_DNS}")
   fi
-     
-  if [ -n "${HOME}" ]; then
-    if [ "${HOME}" == "/localhost" ]; then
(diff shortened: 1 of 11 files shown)


osterman:
Per the stackoverflow link, did you try the simpler path? `C:\Users\%USERNAME%\AppData\Local\lxss\` (`/mnt/c/Users/...`)

rootfs/templates/wrapper
```diff
@@ -89,6 +89,11 @@ function use() {
     if [ "${HOME}" == "/localhost" ]; then
       echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
     else
+      if [[ $(uname -r) =~ Microsoft$ ]]; then
+        WSL_USER_DIR=${WSL_USER_DIR:-$(ls -d /mnt/c/Users/* | head -1)}
```
osterman: Why does this not take into consideration `WSL_USER_NAME`?

rootfs/templates/wrapper
```diff
@@ -89,6 +89,11 @@ function use() {
     if [ "${HOME}" == "/localhost" ]; then
       echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
     else
+      if [[ $(uname -r) =~ Microsoft$ ]]; then
+        WSL_USER_DIR=${WSL_USER_DIR:-$(ls -d /mnt/c/Users/* | head -1)} 
+        WSL_USER_NAME=${WSL_USER_NAME:-user} 
+        HOME="${WSL_USER_DIR}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc/LocalState/rootfs/home/${WSL_USER_NAME}"
```
osterman: > `CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc` 

This part seems highly unique. I hope we can find an alternative approach.

vadim-hleif: I think it is okay. Just name of the company and app. Added regular expression for version and some id(maybe) at the end.

rootfs/templates/wrapper
```diff
@@ -89,6 +89,11 @@ function use() {
     if [ "${HOME}" == "/localhost" ]; then
       echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
     else
+      if [[ $(uname -r) =~ Microsoft$ ]]; then
```
osterman: Let's use the fix here: https://github.com/cloudposse/geodesic/issues/199#issuecomment-409610973

osterman: Let's swap out `sed` with `tr`

rootfs/templates/wrapper
```diff
@@ -84,13 +84,29 @@ function use() {
...
+     
   if [ -n "${HOME}" ]; then
     if [ "${HOME}" == "/localhost" ]; then
       echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
     else
-      echo "# Mounting ${HOME} into container"
-      DOCKER_ARGS+=(--volume=${HOME}:/localhost)
+      if  [[ $(cat /proc/version) =~ Microsoft ]] ; 
+      then
+        windowsUserName=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | tr -d '\r')
```
osterman: Let's use `snake_case` everywhere

rootfs/templates/wrapper
```diff
@@ -84,13 +84,29 @@ function use() {
...
+
+        if [ -d "/mnt/c/Users/${windowsUserName}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc" ];
```
osterman: > CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc
Experience tells me this is probably related to a build number or CPU architecture or both. This won't work. 

Instead, do something like this:
```
for dir in /mnt/c/Users/${windowsUserName}/AppData/Local/Packages/*Ubuntu*on*Windows*; do
  if [ -d "${dir}/LocalState/rootfs${HOME}" ]; then 
     ...
     break
  fi
done
```

rootfs/templates/wrapper
```diff
@@ -84,13 +84,29 @@ function use() {
...
-      echo "# Mounting ${HOME} into container"
-      DOCKER_ARGS+=(--volume=${HOME}:/localhost)
+      if  [[ $(cat /proc/version) =~ Microsoft ]] ; 
+      then
+        windowsUserName=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | tr -d '\r')
+
+        if [ -d "/mnt/c/Users/${windowsUserName}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc" ]; 
+        then 
+          localHome=C:/Users/${windowsUserName}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu18.04onWindows_79rhkp1fndgsc/LocalState/rootfs${HOME}
+        elif [ -d "/mnt/c/Users/${windowsUserName}/AppData/Local/lxss/" ]
```
osterman: In bash, always use the following convention...
```
if [....]; then

fi
```

rootfs/templates/wrapper
```diff
@@ -84,13 +84,29 @@ function use() {
...
+        then 
+          localHome=C:/Users/${windowsUserName}/AppData/Local/lxss${HOME}
+        fi
+
+        echo Detected Windows Subsystem for Linux, mounting $localHome instead of $HOME
```
osterman: Use quotes 
```
echo "...."
```

rootfs/templates/wrapper
```diff
@@ -84,13 +84,29 @@ function use() {
     # Implement DNS fix related to https://github.com/docker/docker/issues/24344
     DOCKER_ARGS+=("--dns=${DOCKER_DNS}")
   fi
-
+
```
osterman: Let's introduce a `LOCAL_HOME` variable. The user can define it, then we use it. If they don't define it, then we do this....

e.g.

```
if [ -z "${LOCAL_HOME}" ]; then
# not defined
...
fi
```

rootfs/templates/wrapper
```diff
@@ -84,13 +84,49 @@ function use() {
...
+          echo "ERROR: can't identify user home directory, you may specify path via LOCAL_HOME variable"
+          exit 1;
+        else
+          echo Detected Windows Subsystem for Linux, mounting $local_home instead of $HOME
+          echo "# Mounting ${local_home} into container"
+          DOCKER_ARGS+=(--volume=${local_home}:/localhost)
+        fi
+      else
+        echo "# Mounting ${HOME} into container"
+        DOCKER_ARGS+=(--volume=${HOME}:/localhost)
```
osterman: `LOCAL_HOME` should not just apply to WSL.

rootfs/templates/wrapper
```diff
@@ -84,13 +84,49 @@ function use() {
...
+     
   if [ -n "${HOME}" ]; then
     if [ "${HOME}" == "/localhost" ]; then
       echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
     else
-      echo "# Mounting ${HOME} into container"
-      DOCKER_ARGS+=(--volume=${HOME}:/localhost)
+      if  [[ $(cat /proc/version) =~ Microsoft ]] ; 
+      then
+        if [ -z "${LOCAL_HOME}" ];
```
osterman: The logic should be like this:

```
if [ -z "${LOCAL_HOME}" ]; then
... now try to figure it out...
if WSL, and we can't find HOME, error. 
otherwise, LOCAL_HOME=${HOME}
fi
...

DOCKER_ARGS+=(--volume=${LOCAL_HOME}:/localhost)

```

rootfs/templates/wrapper
```diff
@@ -84,13 +84,49 @@ function use() {
...
+
+          if [ -d "/mnt/c/Users/${windows_user_name}/AppData/Local/lxss/" ]; 
+          then 
+            local_home=${user_local_app_data}/lxss${HOME}
+          else 
+            for dir in /mnt/c/Users/${windows_user_name}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu* 
+            do
+              if [ -d $dir ]
+              then
+                folder_name=$(echo ${dir} | sed 's:.*/::')
```
osterman: I think what you want here is `$(basename ${dir})`

rootfs/templates/wrapper
```diff
@@ -84,13 +84,49 @@ function use() {
...
+          windows_user_name=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | tr -d '\r')
+          user_local_app_data=$(cmd.exe /c echo %LOCALAPPDATA% | tr -d '\r' | sed -e 's/\\/\//g')
+
+          if [ -d "/mnt/c/Users/${windows_user_name}/AppData/Local/lxss/" ]; 
+          then 
+            local_home=${user_local_app_data}/lxss${HOME}
+          else 
+            for dir in /mnt/c/Users/${windows_user_name}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu* 
+            do
+              if [ -d $dir ]
```
osterman: We dont' want to test if `$dir` exists. We know it exists since we glob matched. What we want to check is that our derived `local_home` path exists.

rootfs/templates/wrapper
```diff
@@ -84,13 +84,49 @@ function use() {
...
+              then
+                folder_name=$(echo ${dir} | sed 's:.*/::')
+                local_home=${user_local_app_data}/Packages/${folder_name}/LocalState/rootfs${HOME}
```
osterman: We shouldn't break until we find one that exists. This is not guaranteed to exist.

rootfs/templates/wrapper
vadim-hleif: But if folder doesn't exist  for get pattern as `$dir`

rootfs/templates/wrapper
```diff
@@ -85,15 +85,38 @@ function use() {
     DOCKER_ARGS+=("--dns=${DOCKER_DNS}")
   fi
 
-  if [ -n "${HOME}" ]; then
-    if [ "${HOME}" == "/localhost" ]; then
-      echo "WARNING: not mounting ${HOME} because it conflicts with geodesic"
+  if [ "${HOME}" == "/localhost" ]; then
```
osterman: This is the wrong place to check for matching.

rootfs/templates/wrapper
```diff
@@ -85,15 +85,38 @@ function use() {
...
+    if [ -z "${local_home}" ]; then
+      echo "ERROR: can't identify user home directory, you may specify path via LOCAL_HOME variable"
+      exit 1;
     else
-      echo "# Mounting ${HOME} into container"
-      DOCKER_ARGS+=(--volume=${HOME}:/localhost)
+      echo Detected Windows Subsystem for Linux, mounting $local_home instead of $HOME
     fi
+    
   fi
```
osterman: This is where we should validate `local_home` is not `/localhost`

rootfs/templates/wrapper
```diff
@@ -85,15 +85,38 @@ function use() {
...
+    local_home=${LOCAL_HOME}
+  elif [[ $(cat /proc/version) =~ Microsoft ]]; then
+    windows_user_name=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | tr -d '\r')
+    user_local_app_data=$(cmd.exe /c echo %LOCALAPPDATA% | tr -d '\r' | sed -e 's/\\/\//g')
+
+    if [ -d "/mnt/c/Users/${windows_user_name}/AppData/Local/lxss/" ]; then 
+      local_home=${user_local_app_data}/lxss${HOME}
+    else 
+      shopt -s nullglob
+      for dir in /mnt/c/Users/${windows_user_name}/AppData/Local/Packages/CanonicalGroupLimited.Ubuntu*
```
osterman: Use `; do` format.

osterman: https://google.github.io/styleguide/shell.xml?showone=Loops#Loops

PR #205, opened 2018-08-01
Add `fluentd-elasticsearch-logs` and `kibana` to `helmfile`

aknysh:
## what
* Add `fluentd-elasticsearch-logs` and `kibana` to `helmfile`
* Add `kibana` to `portal`

## why
* Easily deploy `fluentd` log forwarder for Elasticsearch and `Kibana` for search UI
* Allow accessing `Kibana` from the K8s portal


rootfs/conf/kops/helmfile.yaml
```diff
@@ -989,6 +989,67 @@ releases:
...
+- name: "fluentd-elasticsearch-logs"
+  namespace: "kube-system"
+  labels:
+    chart: "fluentd-kubernetes"
+    component: "elasticsearch"
+    namespace: "kube-system"
+    default: "false"
+  chart: "cloudposse-incubator/fluentd-kubernetes"
+  version: "0.2.0"
+  set:
```
osterman: Use inline values instead

rootfs/conf/kops/helmfile.yaml
```diff
@@ -1146,6 +1207,13 @@ releases:
     - name: "backends.docs.external"
       value: "true"
 
+    - name: "backends.kibana.name"
```
osterman: Because ElasticSearch is $$ I think we should not add it by default. Maybe add a conditional for `PORTAL_BACKEND_KIBANA_ENABLED`

osterman:
Move to `cloudposse/helmfiles/helmfile.d`

aknysh:
moved to https://github.com/cloudposse/helmfiles

PR #207, opened 2018-08-02
Restore white checkmark in unicode prompt

osterman:
## what
* Restore ✅

## why
* The ✔is too discreet and doesn't match ❌


osterman:
Aha, this is probably why tamsky changed it.

image

Causes same-line wrapping, while the other glyph does not.

image

osterman:
Fixed by adding brackets.

image

https://github.com/folti/bash-it/commit/441d0e2d318a002b665cb7617287e953398571c8

PR #210, opened 2018-08-03
Strategically signal bash to resize

osterman:
## what
* Only call `kill -WINCH $$` when dimensions of screen change

## why
* Theory is that it contributes to this error... image
* I suspect it could be related to bash not responding to `kill -WINCH $$` within some kind of deadline

### DIFF
--- a/rootfs/etc/profile.d/prompt.sh
+++ b/rootfs/etc/profile.d/prompt.sh
@@ -2,6 +2,9 @@
 # Allow bash to check the window size to keep prompt with relative to window size
 shopt -s checkwinsize
 
+# Cache the current screen size
+export SCREEN_SIZE="${LINES}x${COLUMNS}"
+
 export PROMPT_COMMAND=prompter
 function prompter() {
     for hook in ${PROMPT_HOOKS[@]}; do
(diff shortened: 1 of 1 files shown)


rootfs/etc/profile.d/prompt.sh
```diff
@@ -58,7 +68,7 @@ function geodesic_prompt() {
     ROLE_PROMPT="(none)"
   fi
 
-  PS1="\["$'${STATUS}'"\]"
+  PS1=$'${STATUS}'
```
osterman: The extra brackets should only be used when ANSI is present. Otherwise, bash cursor position is miscalculated, in this case by 4 characters when pressing up-arrow to cycle through bash history.

image

rootfs/etc/profile.d/prompt.sh
```diff
@@ -11,7 +14,14 @@ function prompter() {
...
 function reload() {
-  kill -WINCH $$
+  local current_screen_size="${LINES}x${COLUMNS}"
+  # Detect changes in screensize
+  if [ "${current_screen_size}" != "${SCREEN_SIZE}" ]; then
+    echo "# Screen resized to ${current_screen_size}"
+    export SCREEN_SIZE=${current_screen_size}
+    # Instruct shell that window size has changed to ensure lines wrap correctly
+    kill -WINCH $$
+  fi
```
tamsky: The more I look at this, the more I don't think any of it should be necessary.
bash should be handling all this when jobs exit, and `checkwinsize` is `on`, which it is.

Issue #209, opened 2018-08-02
Deprecate Helmfile.yaml and `kiam` scripts


osterman:
## what
* Deprecate the `helmfile.yaml`

## why
* Moved to https://github.com/cloudposse/helmfiles

osterman:
@goruha @MoonMoon1919 heads up

PR #213, opened 2018-08-08
Deprecate helmfile.yaml and kiam scripts

alebabai:
## what
deprecate helmfile.yaml and kiam scripts

## why
#209 

## references
- <https://github.com/cloudposse/prod.cloudposse.co/pull/8/files> (example usage)


Dockerfile
```diff
@@ -53,6 +53,11 @@ COPY --from=packages /packages/install/ /packages/install/
 # Copy select binary packages
 COPY --from=packages /dist/ /usr/local/bin/
 
+# Copy helmfiles and kiam scripts
```
aknysh: We probably don't need to copy `helmfiles` into `geodesic` image.
We do it in reference architecture modules like this https://github.com/cloudposse/testing.cloudposse.co/blob/master/Dockerfile#L3
@osterman what do you think?

osterman:
Agree with @aknysh. Let’s remove it but not restore it from helmfiles image.

PR #214, opened 2018-08-09
Support non-default WSL root mount path

Martaver:
## what
* Use regex to look up the correct root mount path from `wsl.conf`.

## why
* Users may use non-standard paths

## testing
* I tested the script on my system, and it works - however, wasn't able to test the whole build.


rootfs/templates/wrapper
```diff
@@ -88,25 +88,40 @@ function use() {
   if [ -n "${LOCAL_HOME}" ]; then
     local_home=${LOCAL_HOME}
   elif [[ $(uname -r) =~ Microsoft$ ]]; then
-    windows_user_name=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | tr -d '\r')
+    
+    wsl_config=`cat /etc/wsl.conf`
+    mount_path_regex="\[automount\](.|\n|\r)*root\s+=\s+(\S+)"
```
vadim-hleif: it wouldn't be work, in case when root property are commented

rootfs/templates/wrapper
```diff
@@ -88,25 +88,40 @@ function use() {
   if [ -n "${LOCAL_HOME}" ]; then
     local_home=${LOCAL_HOME}
   elif [[ $(uname -r) =~ Microsoft$ ]]; then
-    windows_user_name=$(/mnt/c/Windows/System32/cmd.exe /c 'echo %USERNAME%' | tr -d '\r')
+    
+    wsl_config=`cat /etc/wsl.conf`
```
vadim-hleif: use `$()` instead ` `` `
[Why is $(...) preferred over backticks?](http://mywiki.wooledge.org/BashFAQ/082)

rootfs/templates/wrapper
vadim-hleif: Instead `if else` and reading `wsl.conf` you can just get the right path from mounted file systems by file system `drvfs ` type and windows path `C:`
`dirname $(findmnt -S C: -t drvfs -no target)`

Issue #212, opened 2018-08-05
Historical command line editing is broken


Nuru:
## what
Concerning version 0.12.6. Also reproduced in 0.11.0.

Editing historical command lines using emacs keys is broken in certain circumstances. My guess is that it has to do with a divergence between the actual length of the prompt when output versus the length of the prompt when queried by the command line editor.

Reproducing this bug is a little tricky. This seems to work for me, but you might have to try some variations.

1. Enter a long command at the command line:
    ```
    echo top level this is a long command
    ```
1. Press the up-arrow &uparrow; to recall the command to the command line.
1. Press ctrl-A to move the cursor to the beginning of the command line.

Expected: cursor hovers over "e" in echo.
Observed: cursor hovers over "c" in echo.

## why
Not only does this cause difficulty in editing historical command lines, it results in a dangerous situation where the command visible on the command line is not exactly what will be submitted when you hit return.

osterman:
What version are you using?

You can set `PROMPT_STYLE=plain` which will probably eliminate your problems

Nuru:
Using `0.12.6`. Previously using `0.11.0` which did not have this problem.

`PROMPT_STYLE=plain` does not solve the problem.

Commenting out `PROMPT_HOOKS+=("geodesic_prompt")` from `/etc/profile.d/prompt.sh` eliminates the issue (and the nice prompt) but of course does not survive a restart of the container.

Nuru:
Replacing `/etc/profile.d/prompt.sh` with version from `0.9.18` does not solve the problem, nor does unsetting `PROMPT_HOOKS`.

osterman:
This is what I see in iTerm2:

![image](https://user-images.githubusercontent.com/52489/43758379-c9c60840-99d0-11e8-9040-56a15bebced1.png)

Maybe related to my terminal (why it works). 

Can you try just radically simplifying the prompt characters in `/etc/profile.d` and removing all the `tput` statements. 

If that works, we can introduce another prompt format.

Nuru:
I'm using the OS X Terminal. I seem to see some improvement in Terminal if I turn on the "Unicode East Asian Ambiguous characters are wide" option in Terminal -> Preferences -> Advanced, but not complete solution. What do you see when you search for "echo" and then start to edit the line (key sequence `ctrl-R e c h o ctrl-A`)? I see a colon left over from search and the cursor over the space between the colon and the "e" in "echo".

[...]

PR #218, opened 2018-08-17
Add color prompt with single byte width characters

osterman:
## what
* Add new *default* prompt style that uses single byte width characters

## why
* The multi-byte unicode characters causes problems on terminals with line wrapping and broken command history
* Better portability
* Wide-Unicode support requires "Use Unicode Version 9 widths" feature flag enabled in iTerm2

## demo
image

## references
* https://github.com/cloudposse/geodesic/issues/212

### DIFF
--- a/rootfs/etc/profile.d/prompt.sh
+++ b/rootfs/etc/profile.d/prompt.sh
@@ -17,7 +17,7 @@ function reload() {
   local current_screen_size="${LINES}x${COLUMNS}"
   # Detect changes in screensize
   if [ "${current_screen_size}" != "${SCREEN_SIZE}" ]; then
-    echo "# Screen resized to ${current_screen_size}"
+    echo "* Screen resized to ${current_screen_size}"
     export SCREEN_SIZE=${current_screen_size}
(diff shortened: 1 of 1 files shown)


rootfs/etc/profile.d/prompt.sh
```diff
@@ -47,13 +47,22 @@ function geodesic_prompt() {
...
       # unicode
       AWS_VAULT_ACTIVE_MARK=$'\u2705 '      # '✅'
       AWS_VAULT_INACTIVE_MARK=$'\u274C '    # '❌'
       BLACK_RIGHTWARDS_ARROWHEAD=$'\u27A4 ' # '➤', suggest '▶' may be present in more fonts
       BANNER_MARK=$'\u29C9 '                # '⧉'
       ;;
+
+    *)
+      # default
+      AWS_VAULT_ACTIVE_MARK=$' \x01'$(tput bold)$(tput setaf 2)$'\x02\u2713 \x01'$(tput sgr0)$'\x02';    # green bold '✓'
```
tamsky: can we document the need for the `\x01` and `\x02` delimiters?

Issue #195, opened 2018-07-29
Fix Shell Concurrency


osterman:
## what
* `aws-vault: error: Server failed: exit status 2`

## why
* `aws-vault server` should only be run once globally across all shell because it binds to `169.254.169.254`

## references
- <https://github.com/cloudposse/root.cloudposse.co/issues/8#issuecomment-408674720>

osterman:
```
curl -sSL -o /dev/null http://169.254.169.254/latest/meta-data/iam/security-credentials
result=$?
if [ $result -eq 0 ]; then
  echo "Metadata server already running"
fi
```

PR #223, opened 2018-08-17
aws-vault - fix shell concurrency

alebabai:
## what
* Check if aws-vault server is already running before attempting to start

## why
* Can only run once since it attempts to bind to `169.254.169.254`

## references
* #195

### DIFF
--- a/rootfs/etc/profile.d/aws-vault.sh
+++ b/rootfs/etc/profile.d/aws-vault.sh
@@ -12,7 +12,7 @@ else
     curl -sSL -o /dev/null --stderr /dev/null http://169.254.169.254/latest/meta-data/iam/security-credentials
     result=$?
     if [ $result -ne 0 ]; then
-        echo "* Started EC2 metadata service at http://169.254.169.254/latest"
+        echo "* Started EC2 metadata service at $(green http://169.254.169.254/latest)"
         aws-vault server &
         AWS_VAULT_ARGS+=("--server")
      else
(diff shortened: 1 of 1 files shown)


rootfs/etc/profile.d/aws-vault.sh
```diff
@@ -8,10 +8,14 @@ else
...
   ln -sf /localhost/.awsvault ${HOME}
-  if [ "${VAULT_SERVER_ENABLED:-true}" == "true" ]; then
+  curl -sSL -o /dev/null --stderr /dev/null http://169.254.169.254/latest/meta-data/iam/security-credentials
+  result=$?
+  if [ "${VAULT_SERVER_ENABLED:-true}" == "true" ] && [ $result -ne 0 ]; then
     echo "* Started EC2 metadata service at http://169.254.169.254/latest"
     aws-vault server &
     AWS_VAULT_ARGS+=("--server")
+  else
+    echo "* Metadata server already running"
```
osterman: This is not always logically correct

aknysh: yep, should be something like this:

if [ "${VAULT_SERVER_ENABLED:-true}" == "true" ]; then
  curl -sSL -o /dev/null --stderr /dev/null http://169.254.169.254/latest/meta-data/iam/security-credentials
  result=$?
  if [ $result -ne 0 ]; then
      echo "* Started EC2 metadata service at http://169.254.169.254/latest"
      aws-vault server &
      AWS_VAULT_ARGS+=("--server")
   else
      echo "* Metadata server already running"
   fi
fi

osterman:
Rebase

rootfs/etc/profile.d/aws-vault.sh
alebabai: you're right, thanks

Issue #187, opened 2018-07-24
s3fs should not remount filesystem if already mounted


osterman:
## what
`mount -a` should not produce error

## why

s3fs script should check if filesystem is already mounted

PR #227, opened 2018-08-20
s3fs - check if fs is already mounted

alebabai:
## what
check if fs is already mounted before calling `mount -a`

## why
#187

### DIFF
--- a/rootfs/usr/bin/s3fs
+++ b/rootfs/usr/bin/s3fs
@@ -10,4 +10,6 @@ if [ ! -w /dev/fuse ]; then
 fi
 
 # Evaluate the bash envs passed in via args and then execute goofys
-exec goofys --file-mode=0600 --dir-mode=0700 --region=${BUCKET_REGION} --sse --acl=private ${DEVICE##s3://} ${MOUNT_POINT}
+if ! mountpoint -q "${MOUNT_POINT}"; then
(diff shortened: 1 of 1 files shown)


rootfs/usr/local/bin/s3
```diff
@@ -20,7 +20,7 @@ if [[ "$ACTION" = "fstab" ]]; then
 	# Create mount point directory
 	mkdir -p "${MOUNT_POINT}"
 elif [[ "$ACTION" = "mount" ]]; then
-	mount -a
+	if ! mountpoint -q "${MOUNT_POINT}"; then mount -a; fi
```
osterman: Move this fix to s3fs wrapper script.

osterman: (that way `mount -a` won't fail)

alebabai: Did you mean `rootfs/usr/local/bin/init-terraform` script where `s3` command is used?
Or something like this (in file `rootfs/usr/bin/s3fs`):
```bash
if ! mountpoint -q "${MOUNT_POINT}"; then
    exec goofys --file-mode=0600 --dir-mode=0700 --region=${BUCKET_REGION} --sse --acl=private ${DEVICE##s3://} ${MOUNT_POINT}
fi
```

alebabai: it seems to me that the second one is more reasonable..

osterman: Objective is that `mount -a` should not fail. Thus, fix belongs in `rootfs/usr/bin/s3fs` which is called by `mount`

PR #237, opened 2018-08-22
[kops] rbac support

### LINKED ISSUE #228: Enable Optional RBAC in Kops Manifest
## what
* Add env flag to enable/disable RBAC (default to off, for backwards compatibility)

## why
* RBAC is the new norm (for almost a year now)
* We need to start supporting it

alebabai:
## what
* beautify manifest template
* add rbac support (working example from here https://pracucci.com/kubernetes-rbac-with-kops.html)

## why
closes #228


osterman:
@alebabai for reference, one our customers is using:
```
  kubeAPIServer: 
    authorizationMode: RBAC,AlwaysAllow 
    authorizationRbacSuperUser: cluster-admin 
```
Do we need this?

rootfs/templates/kops/default.yaml
```diff
@@ -18,37 +18,45 @@ spec:
   api:
     loadBalancer:
       type: Public
+  {{- if bool (getenv "KOPS_AUTHORIZATION_RBAC_ENABLED") }}
+  kubeAPIServer:
+    authorizationRbacSuperUser: {{ getenv "KOPS_AUTHORIZATION_RBAC_SUPER_USER" | default "admin" }}
+  {{- end }}
   authorization:
+    {{- if bool (getenv "KOPS_AUTHORIZATION_RBAC_ENABLED") }}
+    rbac: {}
```
osterman: See comments

alebabai:
@osterman 
> Parameterize rbac section
what do you mean, what kind parameters?

osterman:
What are the parameters to `rbac: {}`?

rootfs/templates/kops/default.yaml
```diff
@@ -18,37 +18,45 @@ spec:
   api:
     loadBalancer:
       type: Public
+  {{- if bool (getenv "KOPS_AUTHORIZATION_RBAC_ENABLED") }}
+  kubeAPIServer:
```
osterman: Do we need `authorizationMode`?

alebabai: I don't think so, because it doesn't specified in cluster.spec

alebabai:
> What are the parameters to rbac: {}?
it seems to me that there aren't any params here
https://github.com/kubernetes/kops/blob/master/docs/apireference/build/documents/_generated_authorizationspec_v1alpha2_kops_definition.md
https://github.com/kubernetes/kops/blob/master/docs/apireference/build/documents/_generated_rbacauthorizationspec_v1alpha2_kops_definition.md

rootfs/templates/kops/default.yaml
osterman: Bump

alebabai: i read about `kubeAPIServer` props, so the only thing that could be usefull is `authorizationMode: RBAC,AlwaysAllow`. It explicitly declares authorization modes.

rootfs/templates/kops/default.yaml
```diff
@@ -18,37 +18,46 @@ spec:
...
+    authorizationMode: RBAC,AlwaysAllow
```
osterman: make this an env that defaults to `RBAC,AlwaysAllow`

alebabai: done

Issue #241, opened 2018-08-22
[kops] Add Support for Enhanced Monitoring


osterman:
## what
```
detailedInstanceMonitoring: {{ getenv "KOPS_CLOUDWATCH_DETAILED_MONITORING" "true" }}
```

## why
* More detailed monitoring

PR #244, opened 2018-08-23
kops - detailed monitoring support

alebabai:
## what
add detailed monitoring support to IG

## why
#241

### DIFF
--- a/rootfs/templates/kops/default.yaml
+++ b/rootfs/templates/kops/default.yaml
@@ -108,6 +108,7 @@ metadata:
     kops.k8s.io/cluster: {{getenv "KOPS_CLUSTER_NAME"}}
   name: master-{{.}}
 spec:
+  detailedInstanceMonitoring: {{ getenv "KOPS_CLOUDWATCH_DETAILED_MONITORING" "true" }}
   associatePublicIp: false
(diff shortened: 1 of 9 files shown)


rootfs/templates/kops/default.yaml
```diff
@@ -108,6 +108,7 @@ metadata:
     kops.k8s.io/cluster: {{getenv "KOPS_CLUSTER_NAME"}}
   name: master-{{.}}
 spec:
+  detailedInstanceMonitoring: {{ getenv "KOPS_CLOUDWATCH_DETAILED_MONITORING" "true" }}
```
osterman: Lets default to false since it increases cost of operation.

rootfs/templates/kops/default.yaml
```diff
@@ -126,6 +127,7 @@ metadata:
     kops.k8s.io/cluster: {{getenv "KOPS_CLUSTER_NAME"}}
   name: nodes
 spec:
+  detailedInstanceMonitoring: {{ getenv "KOPS_CLOUDWATCH_DETAILED_MONITORING" "true" }}
```
osterman: Default to false

PR #250, opened 2018-08-27
Add MFA Helper Function

osterman:
## what
* Add function that emits TOTP MFA code

## why
* Useful for CI/CD and demos


osterman:
@aknysh I struggle with deciding if this should be a function or a script

osterman:
> We can also consider using chamber here to read the MFA security string (and not put it into a file in the first place)

Catch-22 that would also defeat purpose of MFA in this use-case

Issue #249, opened 2018-08-25
Use TLS for APK repositories


osterman:
## what
* Use TLS for apk repositories
* Use the `sed` hack to use fastly upstream CDN

## why
* More secure
* Mitigate MITM attacks

## references
* https://github.com/gliderlabs/docker-alpine/issues/184

alebabai:
@osterman is it still require some changes or it can be closed?

osterman:
Can close

PR #262, opened 2018-09-04
apk: use https fastly repositories

alebabai:
## what
Switch to the HTTPS alpine apk repositories

## why
#249

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM alpine:3.8 as python 
 
 COPY requirements.txt /requirements.txt
+RUN sed -i 's/http\:\/\/dl-cdn.alpinelinux.org/https\:\/\/alpine.global.ssl.fastly.net/g' /etc/apk/repositories
 RUN apk add python python-dev libffi-dev gcc py-pip py-virtualenv linux-headers musl-dev openssl-dev make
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM alpine:3.8 as python 
 
 COPY requirements.txt /requirements.txt
+RUN sed -i 's/http\:\/\/dl-cdn.alpinelinux.org/https\:\/\/alpine.global.ssl.fastly.net/g' /etc/apk/repositories
```
osterman: Change delimiter to some other character than slash (e.g. use ,). Then you don’t need to escape everything.

Dockerfile
```diff
@@ -35,7 +36,8 @@ USER root
 # oath-toolkit
 # https://www.nongnu.org/oath-toolkit/
 # https://www.nongnu.org/oath-toolkit/oathtool.1.html
-RUN echo "@testing http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
+RUN sed -i 's/http\:\/\/dl-cdn.alpinelinux.org/https\:\/\/alpine.global.ssl.fastly.net/g' /etc/apk/repositories \
```
osterman: Change delimiter.

osterman:
Fix CI

Issue #261, opened 2018-09-04
Shell Formatting


osterman:
## what
* Use `shfmt` to format all shell scripts 
* Consider adding target to `build-harness`

## why
* Improve consistency

Issue #260, opened 2018-09-04
Add Shell Linting


osterman:
## what
* Use `shellcheck` to lint all bash scripts
* Consider adding target to `build-harness`

## why
* Reduce bugs, improve consistency

PR #263, opened 2018-09-04
Add shfmt

aknysh:
## what
* Add `shellcheck` 
* Add `shfmt`
* Bump `GCLOUD_SDK_VERSION`

## why
* `shfmt` to format all shell scripts https://github.com/cloudposse/geodesic/issues/261
* `shellcheck` to lint all bash scripts https://github.com/cloudposse/geodesic/issues/260
* Use latest `GCLOUD_SDK_VERSION`
```
Your current Cloud SDK version is: 179.0.0
The latest available version is: 214.0.0
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.8 as python 
+FROM alpine:3.8 as python
 
 COPY requirements.txt /requirements.txt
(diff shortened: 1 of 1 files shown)


aknysh:
but should not those two packages be in `geodesic` for CI/CD?

osterman:
I am torn. We can add it though if you think that will make writing tests easier.

aknysh:
let's add them (does not hurt anything).

Issue #258, opened 2018-09-04
[kops] Add Support for aws-iam-authenticator


osterman:
## what
* Add support for the IAM authenticator for kubernetes

## why
* RBAC for users

## references
* https://github.com/kubernetes-sigs/aws-iam-authenticator

## depends on 
* https://github.com/cloudposse/packages/issues/64

PR #282, opened 2018-09-18
[kops]: add aws-iam-authenticator support

alebabai:
## what
add aws-iam-authenticator configuration to kops manifest template

## why
#258 
https://github.com/kubernetes-sigs/aws-iam-authenticator#kops-usage

### DIFF
--- a/rootfs/templates/kops/default.yaml
+++ b/rootfs/templates/kops/default.yaml
@@ -31,10 +31,26 @@ spec:
   api:
     loadBalancer:
       type: Public
-  {{- if bool (getenv "KOPS_AUTHORIZATION_RBAC_ENABLED" "false") }}
+  hooks:
+  {{- if bool (getenv "KOPS_AWS_IAM_AUTHENTICATOR_ENABLED" "false") }}
+    - name: kops-hook-authenticator-config.service
+      before:
+        - kubelet.service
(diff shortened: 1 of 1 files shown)


rootfs/templates/kops/default.yaml
```diff
@@ -31,10 +31,26 @@ spec:
   api:
     loadBalancer:
       type: Public
-  {{- if bool (getenv "KOPS_AUTHORIZATION_RBAC_ENABLED" "false") }}
+  hooks:
+  {{- if bool (getenv "KOPS_AWS_IAM_AUTHENTICATOR_ENABLED" "false") }}
+    - name: kops-hook-authenticator-config.service
+      before:
+        - kubelet.service
+      roles: [Master]
```
osterman: Let's use list notation for consistency
```
roles:
  - "Master"
```

rootfs/templates/kops/default.yaml
```diff
@@ -31,10 +31,26 @@ spec:
...
+      manifest: |
```
osterman: Shouldn't it be?
```
manifest: |-
```

alebabai: i picked it from official documentation.
`|-` just removes \n from the end

PR #283, opened 2018-09-19
Update `packages`

aknysh:
## what
* Bump `packages` version

## why
* It has [aws-iam-authenticator](https://github.com/kubernetes-sigs/aws-iam-authenticator) which is used to authenticate to EKS clusters on AWS using IAM
* We will implement the solution for EKS described here https://github.com/terraform-providers/terraform-provider-kubernetes/issues/161

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ FROM google/cloud-sdk:216.0.0-alpine as google-cloud-sdk
 #
 # Cloud Posse Package Distribution
 #
-FROM cloudposse/packages:0.25.0 as packages
+FROM cloudposse/packages:0.26.0 as packages
 
 WORKDIR /packages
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -25,7 +25,7 @@ WORKDIR /packages
 #
 # Repo: <https://github.com/cloudposse/packages>
 #
-ARG PACKAGES="awless aws-vault cfssl cfssljson chamber fetch figurine github-commenter gomplate goofys helm helmfile kops kubectl kubectx kubens sops stern terraform terragrunt yq shellcheck shfmt"
+ARG PACKAGES="awless aws-vault cfssl cfssljson chamber fetch figurine github-commenter gomplate goofys helm helmfile kops kubectl kubectx kubens sops stern terraform terragrunt yq shellcheck shfmt aws-iam-authenticator"
```
osterman: I believe packages are sorted alphabetically

PR #311, opened 2018-11-14
Add configurable `idleTimeoutSeconds` for k8s API load balancer

aknysh:
## what
* Add configurable `idleTimeoutSeconds` for k8s API load balancer

## why
* Some `helm` deployments take more than 5 minutes (default apiserver ELB timeout)
* https://github.com/kubernetes/kops/pull/1886

### DIFF
--- a/rootfs/templates/kops/default.yaml
+++ b/rootfs/templates/kops/default.yaml
@@ -31,7 +31,7 @@ spec:
   api:
     loadBalancer:
       type: Public
-      idleTimeoutSeconds: {{ getenv "KOPS_API_SERVER_ELB_TIMEOUT" "300" }}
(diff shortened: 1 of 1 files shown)


rootfs/templates/kops/default.yaml
```diff
@@ -31,12 +31,13 @@ spec:
   api:
     loadBalancer:
       type: Public
+      idleTimeoutSeconds: {{ getenv "KOPS_API_SERVER_ELB_TIMEOUT" "300" }}
```
osterman:
```suggestion
      idleTimeoutSeconds: {{ getenv "KOPS_API_LOAD_BALANCER_IDLE_TIMEOUT_SECONDS" "300" }}
```

osterman: Also, let's increase to 600 by default

PR #327, opened 2018-12-10
Update Dockerfile

goruha:
## What
* Use kubernetes version 1.10.11 

## Why
* To address security issue https://github.com/kubernetes/kubernetes/issues/71411


goruha:
Duplicates #328

PR #332, opened 2018-12-21
[etc/profile.d] automatically export env to terraform

osterman:
## what
 * automatically export current "start up" environment for use with terraform

## why
* Improve consistency
* we're moving away from requiring users to specify a lot of `TF_VAR_*` envs in the `Dockerfile`
* Too much duplication and difficult to keep them in sync
* Many envs are used not only by terraform, but by other tools


rootfs/etc/profile.d/tfenv.sh
```diff
@@ -0,0 +1,3 @@
+# Automatically export the current environment to `TF_VAR` 
+# Use a regex defined in the `TFENV_WHITELIST` and `TFENV_BLACKLIST` environment variables to exclude variables
```
aknysh:
```suggestion
# Use a regex defined in the `TFENV_WHITELIST` and `TFENV_BLACKLIST` environment variables to include and exclude variables 
```

PR #335, opened 2018-12-31
[direnv] Add whitelist path

osterman:
## what
* Add `direnv` whitelist path of `/conf`
* Use `/etc` as the `XDG_CONFIG_HOME` since it's not a multi-user system

## why
* Move away from *requiring* envs in the `Dockerfile` because 
- (a) it complicates coldstart because the `Dockerfile` needs to be updated & rebuilt
- (b) makes all envs global, while envs might mean different things in different contexts
- (c) makes it hard to know which projects consume the envs since they are defined globally. this leads to technical debt.
- (d) complicates local development b/c ENVs are defined in the `Dockerfile`. It's easier if they are in the `.envrc` b/c these can be updated/used during local development in `/localhost`

## caveats
* `XDG_CONFIG_HOME` was previously defined to `/localhost/.geodesic` which meant configurations might vary between workstations. This risks inconsistencies.


jdn-za:
So one aspect I love about this idea is that it keeps things cleaner and mostly more intuitive.

One aspect I have previously found myself disliking is, unless changed from defaults, direnv can be too "transparent" to users not expecting it. 

Possibly solved by having the zsh/bash shell display something about the direnv that is in effect

osterman:
> Possibly solved by having the zsh/bash shell display something about the direnv that is in effect

Everytime you enter the directory, you'll see a notice like this one:
```
direnv: loading .envrc
direnv: export +AWS_REGION +CONFIGS +TF_VAR_repos_dir
```

osterman:
Plus, `direnv` can be called just like `chamber` (so not automated)
```
direnv exec /some/folder my-command
```

jdn-za:
Awesome

PR #348, opened 2019-01-06
Support from-module

osterman:
## what
* Support `-from-module` argument

## why
* Initialize projects

### DIFF
--- a/rootfs/usr/local/bin/init-terraform
+++ b/rootfs/usr/local/bin/init-terraform
@@ -18,6 +18,7 @@ fi
 	exit 1
 )
 
+[ -z "${TF_MODULE}" ] || TF_ARGS+=("-from-module=${TF_MODULE}")
 [ -z "${TF_BUCKET}" ] || TF_ARGS+=("-backend-config=bucket=${TF_BUCKET}")
 [ -z "${TF_BUCKET_REGION}" ] || TF_ARGS+=("-backend-config=region=${TF_BUCKET_REGION}")
 [ -z "${TF_DYNAMODB_TABLE}" ] || TF_ARGS+=("-backend-config=dynamodb_table=${TF_DYNAMODB_TABLE}")
@@ -26,4 +27,4 @@ fi
 
 s3 mount
 
-exec terraform init ${TF_ARGS[@]}
+exec terraform init ${TF_ARGS[@]} $*


rootfs/usr/local/bin/init-terraform
```diff
@@ -4,26 +4,41 @@
...
 
-s3 mount
+# Mount all filesystems (E.g. `goofys`)
+mount -a
+
+# Initialize terraform state
+terraform init ${TF_ARGS[*]} $*
+
+# Copy the current directory configuration & state to the module
+if [ -d "${TF_DIR}" ] && [ -n "${TF_FILES}" ] && [ -n "${TF_FROM_MODULE}" ]; then
```
aknysh: `TF_FROM_MODULE` not used below?

osterman: The objective is to only execute this code block when using `TF_FROM_MODULE` and when we have files in the current directory (`TF_FILES`) that we need to "overlay" on top of the existing configuration

joshmyers:
Have tested this against master and behaviour LGTM, although it doesn't quite work using the above Make style interface. Example:

```
SERVICE ?= $(notdir $(CURDIR))
TF_FROM_MODULE ?= git::git@github.com:ImpactHealthio/terraform-root-modules.git//aws/$(SERVICE)?ref=tags/0.5.7
TF_BUCKET_PREFIX ?= $(SERVICE)

export

init: clean preflight
        @init-terraform .module

clean:
        @rm -rf .module

plan: init
        @$(MAKE) --no-print-directory -C .module plan

apply: init
        @$(MAKE) --no-print-directory -C .module apply
```

Note that `TF_BUCKET_PREFIX ?= $(SERVICE)` is required in the top level Makefile otherwise https://github.com/cloudposse/geodesic/blob/024d4e4475d3fe5deffa8c0c9a0db38c729a316e/rootfs/usr/local/bin/init-terraform#L6 defaults to `.module` in our case, which is not the right S3 key to use for the state.

Also the second `init-terraform` that gets run as part of the terraform-root-modules Makefile vendored into `.module` is failing because we export `TF_FROM_MODULE` in the top level Makefile for its `init-terraform`, but then we don't want to export `TF_FROM_MODULE` into the vendored Makefile run otherwise it also tried to run from a module. Example:

Given a top level Makefile as above:

```
✓   (joany-staging-admin) cloudtrail ⨠  make plan
Copying configuration from "git::git@github.com:ImpactHealthio/terraform-root-modules.git//aws/cloudtrail?ref=tags/0.5.7"...
Initializing modules...
- module.cloudtrail
  Getting source "git::https://github.com/cloudposse/terraform-aws-cloudtrail.git?ref=tags/0.3.0"
- module.cloudtrail.cloudtrail_label
  Getting source "git::https://github.com/cloudposse/terraform-null-label.git?ref=tags/0.3.3"

Initializing the backend...

Successfully configured the backend "s3"! Terraform will automatically
use this backend unless the backend configuration changes.

Initializing provider plugins...
- Checking for available provider plugins on https://releases.hashicorp.com...
- Downloading plugin for provider "aws" (1.54.0)...
- Downloading plugin for provider "null" (1.0.0)...

The following providers do not have any version constraints in configuration,
so the latest version was installed.

To prevent automatic upgrades to new major versions that may contain breaking
changes, it is recommended to add version = "..." constraints to the
corresponding provider blocks in configuration, with the constraint strings
suggested below.

* provider.aws: version = "~> 1.54"
* provider.null: version = "~> 1.0"

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.

The working directory already contains files. The -from-module option requires
an empty directory into which a copy of the referenced module will be placed.

To initialize the configuration already in this working directory, omit the
-from-module option.
make[1]: *** [Makefile:5: init] Error 1
make: *** [Makefile:24: plan] Error 2
```

joshmyers:
One other workaround could be:

```
SERVICE ?= $(notdir $(CURDIR))
TF_ROOT_MODULE_VERSION ?= 0.5.7
TF_BUCKET_PREFIX ?= $(SERVICE)

export

init: clean preflight
        @TF_FROM_MODULE="git::git@github.com:ImpactHealthio/terraform-root-modules.git//aws/$(SERVICE)?ref=tags/$(TF_ROOT_MODULE_VERSION)" init-terraform .module

clean:
        @rm -rf .module

plan: init
        @$(MAKE) --no-print-directory -C .module plan

apply: init
        @$(MAKE) --no-print-directory -C .module apply
 ```

osterman:
> Clearly we don't want the second terraform init to be using the `-from-module`

That's a good point.

osterman:
> defaults to .module in our case, which is not the right S3 key to use for the state

another good catch

PR #349, opened 2019-01-07
Set TF_PLUGIN_CACHE_DIR at runtime

joshmyers:
## what

Move `TF_PLUGIN_CACHE_DIR` to runtime var

## why

This should be a non breaking change as it defaults to 
`/localhost/.terraform/plugins` but allows us to override if necessary
e.g. The atlantis user does not have write access to /localhost

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -189,11 +189,6 @@ ENV KUBE_PS1_VERSION 0.6.0
 ADD https://raw.githubusercontent.com/jonmosco/kube-ps1/${KUBE_PS1_VERSION}/kube-ps1.sh /etc/profile.d/prompt:kube-ps1.sh
 
 #
-# Terraform defaults
-#
-ENV TF_PLUGIN_CACHE_DIR=/localhost/.terraform.d/plugins
-
-#
 # AWS
 #
 ENV AWS_DATA_PATH=/localhost/.aws/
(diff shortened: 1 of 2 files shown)


rootfs/usr/local/bin/init-terraform
```diff
@@ -8,6 +8,9 @@ TF_FILE=${TF_FILE:-terraform.tfstate}
 TF_FILES=(*.tf)
 TF_ARGS=("-backend=true" "-backend-config=key=${TF_BUCKET_PREFIX}/${TF_FILE}")
 
+# Set default plugin cache dir
+export TF_PLUGIN_CACHE_DIR="${TF_PLUGIN_CACHE_DIR:-/localhost/.terraform.d/plugins}"
```
osterman: I don't think this is sufficient. This will work when you call `init-terraform` (which calls `terraform init`), but when you call `terraform plan` or `terraform apply`, will it know where to find them?

osterman: Would this be a good spot? https://github.com/cloudposse/geodesic/blob/master/rootfs/etc/profile.d/terraform.sh

joshmyers: hmm, good point. `/etc/profile.d/terraform.sh` isn't going to work when the Atlantis user has `/bin/false` shell

osterman:
So you'll set `TF_PLUGIN_CACHE_DIR` in `atlantis` then to some path other than `/localhost`? (e.g. `/tmp`)

joshmyers:
@osterman Correct, I'll override if being run via Atlantis

PR #368, opened 2019-01-30
Add okta support

osterman:
## what
* Add support for `aws-okta` by segentio (requires `apk add aws-okta@cloudposse` to work)

## why
* Support alternative ways to assume role


rootfs/etc/profile.d/banner.sh
```diff
@@ -4,7 +4,7 @@ BANNER_COLOR="${BANNER_COLOR:-[36m}"
 BANNER_INDENT="${BANNER_INDENT:-    }"
 BANNER_FONT="${BANNER_FONT:-Nancyj.flf}"
 
-if [ -z "${AWS_VAULT}" ]; then
+if [ -z "${ASSUME_ROLE}" ]; then
 	# Display a banner message for interactive shells (if we're not in aws-vault)
```
aknysh:
```suggestion
	# Display a banner message for interactive shells (if we're not in aws-vault or aws-okta)
```

PR #381, opened 2019-02-22
[prompt]: add hints so readline correctly tracks cursor

tamsky:
## what's changing

- Adds delimiters around terminal escape sequences, so `readline` has accurate cursor position.
Tested for `PROMPT_STYLE=plain` on `xterm`.

- Removes an extraneous double space after the final `=>` arrow in the prompt.

## why

bash->readline was incorrectly wrapping command line text while typing interactively and while browsing shell history.

### DIFF
--- a/rootfs/etc/profile.d/prompt.sh
+++ b/rootfs/etc/profile.d/prompt.sh
@@ -31,7 +31,8 @@ function geodesic_prompt() {
 	case $PROMPT_STYLE in
 	plain)
 		# 8859-1 codepoints:
-		ASSUME_ROLE_ACTIVE_MARK=$(tput bold)$(tput setab 2)$'»'$(tput sgr0)' ' # green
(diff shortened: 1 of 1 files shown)


rootfs/etc/profile.d/prompt.sh
```diff
@@ -31,7 +31,8 @@ function geodesic_prompt() {
 	case $PROMPT_STYLE in
 	plain)
 		# 8859-1 codepoints:
-		ASSUME_ROLE_ACTIVE_MARK=$(tput bold)$(tput setab 2)$'»'$(tput sgr0)' ' # green
+		# '\[' and '\]' are bash prompt delimters around non-printing characters
```
osterman:
```suggestion
		# '\[' and '\]' are bash prompt delimiters around non-printing characters
```

tamsky:
Updated to address review comments.

PR #383, opened 2019-02-27
Rename aws-vault server enabled flag

osterman:
## what
* `$0` does not emit a helpful 
* Add `sudo` so the `--server` mode flag works without the `aws-vault server&` hack
* Update `aws-cli`
* Normalize `AWS_VAULT_SERVER_ENABLED` env
* Add support for `AWS_VAULT_SESSION_TTL`

## why
* `aws-vault` appears to be having some trouble with `--server` mode; rather than rip it out entirely, we'll disable it by default for now

## references
* https://github.com/99designs/aws-vault/issues/339
* https://github.com/99designs/aws-vault/issues/338
* https://github.com/99designs/aws-vault/issues/335


tamsky:
/me wonders if there's a reproduce-able sample testcase where `--server` fails.

PR #387, opened 2019-02-27
profile.d: copy and use inputrc from localhost

tamsky:
## what

adds a small profile.d script that copies the user's homedir `.inputrc` if it exists at geodesic container startup.

## why

Readline keybindings are typically stored in a user's `~/.intputrc`.
This script allows those keybindings to smoothly flow into the geodesic env.

### DIFF
--- a/rootfs/etc/profile.d/inputrc.sh
+++ b/rootfs/etc/profile.d/inputrc.sh
@@ -1,7 +1,4 @@
-# If we can, and do not have one, copy inputrc from localhost.
-[ -f ~/.inputrc ] && [ -f /localhost/.inputrc ] || {
-	cp /localhost/.inputrc ~/.inputrc &&
-		# Use new bindings for current shell.
-		bind -f ~/.inputrc
-}
-# Subsequent shells automatically read ${INPUTRC-"~/.inputrc"} at startup.
+# If we do not have one of our own, source inputrc from localhost.
+[ ! -f ~/.inputrc ] && [ -f /localhost/.inputrc ] &&
+	# Use new bindings for current shell.
+	bind -f /localhost/.inputrc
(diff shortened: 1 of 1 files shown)


rootfs/etc/profile.d/inputrc.sh
```diff
@@ -0,0 +1,7 @@
+# If we can, and do not have one, copy inputrc from localhost.
+[ -f ~/.inputrc ] && [ -f /localhost/.inputrc ] || {
```
osterman: Don't you mean this?

```suggestion
[ ! -f ~/.inputrc ] && [ -f /localhost/.inputrc ] || {
```

rootfs/etc/profile.d/inputrc.sh
```diff
@@ -0,0 +1,7 @@
...
+	cp /localhost/.inputrc ~/.inputrc &&
```
osterman: Why the copy and not just bind?

osterman: Any thoughts on symlinking instead?

tamsky: I've implemented your "just bind" suggestion in the next commit.

rootfs/etc/profile.d/inputrc.sh
tamsky: Yes.  Thanks!

rootfs/etc/profile.d/inputrc.sh
```diff
@@ -0,0 +1,4 @@
+# If we do not have one of our own, source inputrc from localhost.
+[ ! -f ~/.inputrc ] && [ -f /localhost/.inputrc ] &&
```
osterman: Prefer to use `if [ ... ]; then` syntax than long chaining commands.

tamsky: Updated.

PR #388, opened 2019-02-28
Intelligently sync the clocks when they are out of sync

Nuru:
## what
When the System Time and the Hardware Clock disagree, sync them.

Specifically, if one clock is behind the other by more than 10 seconds, set the lagging clock to the leading clock, because it is almost always the case that the slow clock is wrong.

## why
AWS credentials have very limited lifetime which is enforced by tying their validity to real-world time. If the clock is wrong, then authentication will fail, so accurate real-world time is essential.


osterman:
@Nuru looks like maybe some bash formatting is due.

PR #389, opened 2019-02-28
[.inputrc] Use symlink in /conf instead of another script

tamsky:
## What's changing

Removes the script `profile.d/inputrc.sh`.
In its place, we add symlink to the default geodesic user's home directory:
   `/conf/.inputrc -> /localhost/.inputrc` 

This seems simpler than a script that tests for and reads the file from `/localhost`.

## Why

Why add a script when a simple symlink suffices?

### DIFF
--- a/rootfs/conf/.inputrc
+++ b/rootfs/conf/.inputrc
@@ -1,0 +1,1 @@
+/localhost/.inputrc
--- a/rootfs/etc/profile.d/inputrc.sh
+++ b/rootfs/etc/profile.d/inputrc.sh
@@ -1,5 +1,0 @@
-# If we do not have one of our own, source inputrc from localhost.
-if [ ! -f ~/.inputrc ] && [ -f /localhost/.inputrc ]; then
-	# Use new bindings for current shell.
-	bind -f /localhost/.inputrc
-fi


osterman:
Since this is a runtime requirement for interactive shells, we've been doing this: 

https://github.com/cloudposse/geodesic/blob/master/rootfs/etc/profile.d/aws.sh


So maybe create an `inputrc.sh` and set it up in there.

tamsky:
> So maybe create an `inputrc.sh` and set it up in there.

An `inputrc.sh` file already exists in `profile.d`.
This PR removes it in favor of a very simple solution: a filesystem symlink to `/localhost/.intputrc`.

My thinking:
"Why add shell logic to decide when to read the file,
 when `/bin/bash` already implements this logic."

When bash starts an interactive shell, it tries to read $INPUTRC or `~/.inputrc` if the file exists.

osterman:
@tamsky what happens if the user does not have an `.inputrc` and we have a symlink to a missing file? Do we see any errors/warnings on the screen?

tamsky:
> what happens if the user does not have an .inputrc and we have a symlink to a missing file? Do we see any errors/warnings on the screen?

@osterman : output from `bash` is identical
-- it does not generate any errors/warnings if the symlink does not point to a readable file.

```
# make run
# Mounting /Users/tamsky into container
# Starting new geodesic session from cloudposse/geodesic:dev
# Exposing port 31706
                                     dP                   oo          
                                     88                               
    .d8888b. .d8888b. .d8888b. .d888b88 .d8888b. .d8888b. dP .d8888b. 
    88'  `88 88ooood8 88'  `88 88'  `88 88ooood8 Y8ooooo. 88 88'  `"" 
    88.  .88 88.  ... 88.  .88 88.  .88 88.  ...       88 88 88.  ... 
    `8888P88 `88888P' `88888P' `88888P8 `88888P' `88888P' dP `88888P' 
         .88                                                          
     d8888P                                                           

IMPORTANT:
* Your $HOME directory has been mounted to `/localhost`
* Use `aws-vault` to manage your sessions
* Run `assume-role` to start a session



| Documentation  | https://docs.cloudposse.com   | Check out documention              |
| Public Slack   | https://slack.cloudposse.com  | Active & friendly DevOps community |
| Paid Support   | hello@cloudposse.com          | Get help fast from the experts     |

Add your local private SSH key to the key chain. Hit ^C to skip.
Enter passphrase for /localhost/.ssh/id_rsa: 
-> Run 'assume-role' to login to AWS with aws-vault
 §  geodesic 
·   (none) ~ => ls -l ~/.inputrc
lrwxrwxrwx 1 root root 19 Mar  8 08:28 /conf/.inputrc -> /localhost/.inputrc
direnv: loading .envrc
-> Run 'assume-role' to login to AWS with aws-vault
 §  geodesic 
·   (none) ~ => cat ~/.inputrc
cat: /conf/.inputrc: No such file or directory
-> Run 'assume-role' to login to AWS with aws-vault
 §  geodesic 
```

PR #390, opened 2019-03-01
[fzf] Fix for non-interactive mode

osterman:
## what
* Disable bash completion for `fzf` if non-interactive terminal

## why
* The upstream bash completion script has a bug where it doesn't handle this use-case


rootfs/etc/profile.d/fzf.sh
```diff
@@ -0,0 +1,5 @@
+# Install fzf shell completion when an interactive shell is enabled
+# This is a fix for: `/etc/bash_completion.d/fzf.sh: line 34: bind: warning: line editing not enabled`
+if [ -t 1 ]; then
```
tamsky: Is this our standard test method for detecting interactive shells?

I'm used to `$PS1` or `$-`

- https://www.tldp.org/LDP/abs/html/intandnonint.html

PR #392, opened 2019-03-01
[fzf] Create symlink conditionally

osterman:
## what
* Fix error when attempting to create link twice

## why
* Entering into a subshell (e.g. with `assume-role`) would emit an error about the target already existing

### DIFF
--- a/rootfs/etc/profile.d/fzf.sh
+++ b/rootfs/etc/profile.d/fzf.sh
@@ -1,5 +1,5 @@
 # Install fzf shell completion when an interactive shell is enabled
 # This is a fix for: `/etc/bash_completion.d/fzf.sh: line 34: bind: warning: line editing not enabled`
-if [ -t 1 ]; then
+if [ -t 1 ] && ! [ -e '/etc/bash_completion.d/fzf.sh' ]; then
 	ln -s /usr/share/bash-completion/completions/fzf /etc/bash_completion.d/fzf.sh
 fi


Nuru:
Why are we doing this instead of using `bash complete`?

osterman:
Can you elaborate?

Nuru:
If we use
```
complete -C '/usr/share/bash-completion/completions/fzf/fzf.sh' fzf
```
as we did [here](https://github.com/cloudposse/geodesic/blob/master/rootfs/etc/profile.d/aws.sh#L17-L20) it won't be a problem to run it twice. So if that works (and maybe it doesn't), I'd prefer that to linking only because it does not have the extra linking problem.

osterman:
So my thinking was to leverage the existing `rc.d` config structure; that said, this might be a better way to do it.

osterman:
> 
as we did [here](https://github.com/cloudposse/geodesic/blob/master/rootfs/etc/profile.d/aws.sh#L17-L20)

@Nuru good point. I'll update it.

Nuru:
I'm in favor of `rc.d` structure when the installation is not conditional, but if it is going to be conditional then I prefer to set it in the shell rather than the file system.

osterman:
> `complete -C '/usr/share/bash-completion/completions/fzf' fzf`

This works, but looks like there's some bigger issue. 

```
/usr/share/bash-completion/completions/fzf: line 34: bind: warning: line editing not enabled
/usr/share/bash-completion/completions/fzf: line 34: bind: warning: line editing not enabled
/usr/share/bash-completion/completions/fzf: line 34: bind: warning: line editing not enabled
/usr/share/bash-completion/completions/fzf: line 34: bind: warning: line editing not enabled

```

PR #395, opened 2019-03-03
Include $STAGE in role prefix

Nuru:
## what
change the starting point for role names (when assuming a role) from $NAMESPACE to $NAMESPACE-$STAGE

## why
Too many roles under just $NAMESPACE, plus roles should be scoped to stages anyway

### DIFF
--- a/rootfs/etc/profile.d/aws-vault.sh
+++ b/rootfs/etc/profile.d/aws-vault.sh
@@ -51,7 +51,7 @@ if [ "${AWS_VAULT_ENABLED:-true}" == "true" ]; then
 				--reverse \
 				--prompt='-> ' \
 				--header 'Select AWS profile' \
-				--query "${AWS_ORG:-${NAMESPACE}}-" \
+				--query "${ASSUME_ROLE_INTERACTIVE_PREFIX:-${NAMESPACE}-${STAGE}-}" \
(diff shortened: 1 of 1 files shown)


rootfs/etc/profile.d/aws-vault.sh
```diff
@@ -51,7 +51,7 @@ if [ "${AWS_VAULT_ENABLED:-true}" == "true" ]; then
 				--reverse \
 				--prompt='-> ' \
 				--header 'Select AWS profile' \
-				--query "${AWS_ORG:-${NAMESPACE}}-" \
+				--query "${ASSUME_ROLE_INTERACTIVE_PREFIX:-${NAMESPACE}-${STAGE}-}" \
```
osterman:
```suggestion
				--query "${ASSUME_ROLE_QUERY:-${NAMESPACE}-${STAGE}-}" \
```

What about calling it `QUERY` since that's what `fzf` calls it?

osterman: (it's also shorter)

Nuru: The ENV var needs to make sense in the global geodesic scope. I want to include `INTERACTIVE` because it only applies to the interactive role chooser

PR #396, opened 2019-03-08
[aws/kops] Mitigate cve-2019-5736 by making runc immutable

Nuru:
## what
Mitigate [cve-2019-5736](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-5736) by making runc immutable

## why
When running a process as root (UID 0) inside a container, that process can exploit a bug in `runc` to gain root privileges on the host running the container. This then allows them unlimited access to the server as well as any other containers on that server. The implemented change mitigates this risk.

## Further information

This mitigation was described [here](https://seclists.org/oss-sec/2019/q1/134) and affirmed [here](https://twitter.com/_staaldraad/status/1095354945073754112) and [here](https://github.com/kubernetes/kops/blob/master/docs/advisories/cve_2019_5736.md#cve-2019-5736---runc-container-breakout) with the caveat that privileged pods (that are effectively root anyway) and pods that have explicitly been granted CAP_LINUX_IMMUTABLE in the securityContext.capabilities remain vulnerable. 

This mitigation is essentially the same as what has been implemented in `kops` version 1.11.1. The real fix is a newer version of Docker that does not have this vulnerability, but none of the released versions of `kops` are certified to run with the fixed versions of Docker. It appears we will have to wait until `kops` version 1.12 to get the fixed version of Docker. 

Bastions are special cases. The bad news is that [hooks on bastions not honored](https://github.com/kubernetes/kops/pull/5856), so this change does not get applied to the bastion host. The good news is that the bastion host does not run anything other containers, so it is protected that way.

### DIFF
--- a/rootfs/templates/kops/default.yaml
+++ b/rootfs/templates/kops/default.yaml
@@ -34,7 +34,8 @@ spec:
       idleTimeoutSeconds: {{ getenv "KOPS_API_LOAD_BALANCER_IDLE_TIMEOUT_SECONDS" "600" }}
   kubeAPIServer:
   {{- if bool (getenv "KOPS_AUTHORIZATION_RBAC_ENABLED" "false") }}
-    authorizationMode: {{ getenv "KOPS_KUBE_API_SERVER_AUTHORIZATION_MODE" "RBAC,AlwaysAllow" }}
+    anonymousAuth: false
+    authorizationMode: {{ getenv "KOPS_KUBE_API_SERVER_AUTHORIZATION_MODE" "RBAC" }}
     authorizationRbacSuperUser: {{ getenv "KOPS_KUBE_API_SERVER_AUTHORIZATION_RBAC_SUPER_USER" "admin" }}
   {{- end }}
   {{- if bool (getenv "KOPS_ADMISSION_CONTROL_ENABLED" "true") }}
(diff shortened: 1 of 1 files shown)


rootfs/templates/kops/default.yaml
```diff
@@ -50,6 +51,8 @@ spec:
     - ResourceQuota
     - NodeRestriction
     - Priority
+    - Initializers
+    - DenyEscalatingExec
```
osterman: I suspect this might cause problems with the teleport node installation via helm and the ability to call `systemctl restart`

rootfs/templates/kops/default.yaml
```diff
@@ -34,7 +34,8 @@ spec:
       idleTimeoutSeconds: {{ getenv "KOPS_API_LOAD_BALANCER_IDLE_TIMEOUT_SECONDS" "600" }}
   kubeAPIServer:
   {{- if bool (getenv "KOPS_AUTHORIZATION_RBAC_ENABLED" "false") }}
-    authorizationMode: {{ getenv "KOPS_KUBE_API_SERVER_AUTHORIZATION_MODE" "RBAC,AlwaysAllow" }}
+    anonymousAuth: false
```
osterman: I think this should be moved above the conditional here as `anonymousAuth` should always be denied, no?

Nuru: In general, with RBAC off, our configuration is `AlwaysAllow`, so I do not think `anonymousAuth` has any effect. I prefer to keep it in the conditional to highlight that.

rootfs/templates/kops/default.yaml
Nuru: It might cause problems, but I prefer to put it in and see if we can find a better solution to those problems.

PR #397, opened 2019-03-10
[aws-vault] Resume current role if aws-vault server is running

Nuru:
## what
If the login shell detects that the `aws-vault` credential server is running, immediately assume that role.

## why
The `aws-vault` credential server is not just convenient, it is essential for enabling long-running tasks (such as `kops rolling-update cluster`) to complete without credentials expiring. However, `aws-vault` has a limitation that `aws-vault exec` will not use an existing credential server, forcing the user to use non-server mode, where expired credentials will not be renewed.  This change works around that limitation by detecting a running credential server and acting as if `assume-role` had been called to assume the role being served by the credential server.

### DIFF
--- a/rootfs/etc/profile.d/aws-vault.sh
+++ b/rootfs/etc/profile.d/aws-vault.sh
@@ -1,10 +1,39 @@
 #!/bin/bash
 
+function assume_active_role() {
+	if [ "${AWS_VAULT_SERVER_ENABLED:-true}" != "true" ]; then
+		return 0
+	fi
+
+	local aws_def_prof="$AWS_DEFAULT_PROFILE"
+	unset AWS_DEFAULT_PROFILE
+
+	curl -sSL --connect-timeout 0.1 --fail -o /dev/null --stderr /dev/null 'http://169.254.169.254/latest/meta-data/iam/security-credentials/local-credentials'
+	if [ $? ]; then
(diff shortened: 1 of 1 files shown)


rootfs/etc/profile.d/aws-vault.sh
```diff
@@ -67,8 +95,16 @@ if [ "${AWS_VAULT_ENABLED:-true}" == "true" ]; then
 	function aws_vault_assume_role() {
 		# Do not allow nested roles
 		if [ -n "${AWS_VAULT}" ]; then
-			echo "Type '$(green exit)' before attempting to assume another role"
-			return 1
+			if [ "$SHLVL" -eq 1 ] && [ "${AWS_VAULT_SERVER_ENABLED:-true}" == "true" ]; then
+				local aws_vault="$AWS_VAULT"
+				local aws_vault_server_enabled="${AWS_VAULT_SERVER_ENABLED:-true}"
+				trap 'export AWS_VAULT="$aws_vault" && export AWS_VAULT_SERVER_ENABLED="$aws_vault_server_enabled" && trap - RETURN' RETURN
```
osterman: Can you add a shell comment above this that explains the intent of what this code does? E.g. why the trap and nested trap?

Nuru: Done

PR #404, opened 2019-03-12
Fix path to small-logo

osulli:
# What

- Removes `docs/` from path to small logo

# Why

- Path was invalid so image was broken


osterman:
Can you confirm `make readme` is not affected?

osulli:
I can `make readme` and all these unexpected changes appear @osterman

osulli:
I understand the changes now and why they came about.
I also fixed an issue where even though the relative path was correct, since it was being imported into another .md file the relativity was wrong... so I just use the URL now.

@osterman

osterman:
Thanks @osulli !

osterman:
@osulli looks like we have an issue with one of our upstream packages related to some recent work to parallelize builds in `cloudposse/packages`. I'll get that fixed.

PR #415, opened 2019-03-20
[wrapper, aws-vault] quickfixes

Nuru:
## what
- [wrapper] mitigate maoOS `bash` bug that causes a failure to find running containers
- [aws-vault] more accurate messaging about EC2 metadata server

## why
- [wrapper] bug fix
- [aws-vault] Previous status message was inaccurate


rootfs/templates/wrapper
```diff
@@ -139,7 +139,10 @@ function use() {
 		--name "${DOCKER_NAME}"
 		--rm
 		--env KUBERNETES_API_PORT=${GEODESIC_PORT})
-	set -o pipefail
+	# With set -o pipefail, the pipe below fails ~50% of the time on Mac due to a bug in bash.
+	# With set +o pipefail, this bug is mitigated, and the only difference is the behavior when `docker ps` fails.
+	# If `docker ps` fails, then it is likely that any further docker action will fail too, so it does not matter.
+	set +o pipefail
```
goruha: why `set -o pipefail` changed to `set +o pipefail`?

Nuru: `	# With set -o pipefail, the pipe below fails ~50% of the time on Mac due to a bug in bash.`

rootfs/etc/profile.d/aws-vault.sh
```diff
@@ -58,10 +58,10 @@ if [ "${AWS_VAULT_ENABLED:-true}" == "true" ]; then
 			curl -sSL --connect-timeout 0.1 -o /dev/null --stderr /dev/null http://169.254.169.254/latest/meta-data/iam/security-credentials
 			result=$?
 			if [ $result -ne 0 ]; then
-				echo "* Started EC2 metadata service at $(green http://169.254.169.254/latest)"
+				echo "* $(green assume-role) will start EC2 metadata service at $(green http://169.254.169.254/latest)"
 				AWS_VAULT_ARGS+=("--server")
 			else
-				echo "* EC2 metadata server already running"
+				echo "* $(red EC2 metadata server already running)"
```
osterman: do we have `yellow`?  This sounds more like a warning

osterman: Never mind, we can keep it red

PR #416, opened 2019-03-20
[rootfs/templates/wrapper] Use docker native filtering with ps

osterman:
## what
* Update wrapper script to use the `docker ps --filter` option rather than pipe/grep pattern

## why
- #415 points out that `pipefail` is unreliable, so lets just eliminate it as a factor in this case

### DIFF
--- a/rootfs/templates/wrapper
+++ b/rootfs/templates/wrapper
@@ -139,13 +139,10 @@ function use() {
 		--name "${DOCKER_NAME}"
 		--rm
 		--env KUBERNETES_API_PORT=${GEODESIC_PORT})
-	# With set -o pipefail, the pipe below fails ~50% of the time on Mac due to a bug in bash.
-	# With set +o pipefail, this bug is mitigated, and the only difference is the behavior when `docker ps` fails.
-	# If `docker ps` fails, then it is likely that any further docker action will fail too, so it does not matter.
-	set +o pipefail
-	docker ps | grep -q ${DOCKER_NAME} >/dev/null 2>&1
-	if [ $? -eq 0 ]; then
-		echo "# Attaching to existing ${DOCKER_NAME} session"
+
+	local container_id=$(docker ps --quiet --filter name="${DOCKER_NAME}" --format "{{`{{ .ID }}`}}")
+	if [ -n "$container_id" ]; then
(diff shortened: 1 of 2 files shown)


rootfs/templates/wrapper
```diff
@@ -139,13 +139,10 @@ function use() {
...
-	# With set -o pipefail, the pipe below fails ~50% of the time on Mac due to a bug in bash.
-	# With set +o pipefail, this bug is mitigated, and the only difference is the behavior when `docker ps` fails.
-	# If `docker ps` fails, then it is likely that any further docker action will fail too, so it does not matter.
-	set +o pipefail
-	docker ps | grep -q ${DOCKER_NAME} >/dev/null 2>&1
-	if [ $? -eq 0 ]; then
-		echo "# Attaching to existing ${DOCKER_NAME} session"
+
+	local container_id=$(docker ps --quiet --filter name="${DOCKER_NAME}" --format "{{`{{ .ID }}`}}")
+	if [ -n "$container_id" ]; then
```
Nuru: This gets around the `pipefail` problem, but perpetuates another problem, such as if one container's name is a substring of another's it will match (`men.cloudposse.co` will match `women.cloudposse.co`). Is there a more robust way to check for this container being run?

osterman: Updated

rootfs/templates/wrapper
```diff
@@ -139,13 +139,10 @@ function use() {
...
 		--env KUBERNETES_API_PORT=${GEODESIC_PORT})
-	# With set -o pipefail, the pipe below fails ~50% of the time on Mac due to a bug in bash.
-	# With set +o pipefail, this bug is mitigated, and the only difference is the behavior when `docker ps` fails.
-	# If `docker ps` fails, then it is likely that any further docker action will fail too, so it does not matter.
-	set +o pipefail
-	docker ps | grep -q ${DOCKER_NAME} >/dev/null 2>&1
-	if [ $? -eq 0 ]; then
-		echo "# Attaching to existing ${DOCKER_NAME} session"
+
+	local container_id=$(docker ps --quiet -a --filter name="^/${DOCKER_NAME}\$" --format '{{`{{ .ID }}`}}')
```
Nuru:
```suggestion
        # the extra curly braces around .ID are because this file goes through go template substitution before being installed as a shell script
```

PR #418, opened 2019-03-22
[direnv] Avoid infinite loop

Nuru:
## what
Fix to `use envrc` addition to `direnv` to avoid an infinite loop condition.

## why
Apparently, direnv executes a separate shell with `nullglob` and `failglob` both off, causing the raw glob to be passed to a future `eval`, causing a loop.

### DIFF
--- a/rootfs/etc/direnv/rc.d/envrc
+++ b/rootfs/etc/direnv/rc.d/envrc
@@ -8,7 +8,7 @@ function use_envrc() {
 	# Export all .envrc files in the current directory
 	
 	for file in $wildcard; do
-		[ -e "$wildcard" ] || break
+		[ -e "$file" ] || break
 		source_env $file
 	done
 }
--- a/rootfs/etc/direnv/rc.d/envrc
+++ b/rootfs/etc/direnv/rc.d/envrc
@@ -5,9 +5,10 @@
 #
 function use_envrc() {
 	local wildcard="${1:-*.envrc}"
-	# Export terraform environent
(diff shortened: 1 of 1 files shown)


tamsky:
> Apparently, direnv executes a separate shell with `nullglob` and `failglob` both off, causing the raw glob to be passed to a future `eval`, causing a loop.

If that is indeed the root cause -- have we considered simply enabling those `shopt`s while this function executes?

Is the current use of `break` appropriate?

This code seems to me to be more intuitive:
```diff
	for file in $wildcard; do
-		source_env $file
+		[ -r "$file" ] && source_env "$file"
	done

**Comment by Nuru** [comment]
> have we considered simply enabling those `shopt`s while this function executes?

I considered it, but 1) it won't work and 2) it would change the execution environment of `source_env`. It won't work because the function is not running in a normal shell, it is running in `direnv` which apparently has parts of `bash` baked in.  

There is nothing terribly wrong with 
```diff
+		[ -r "$file" ] && source_env "$file"
```
although it will suppress errors about non-readable files. The situation we are checking for is one where the `$wildcard` glob fails to match any files and "$file" becomes the glob itself. As an idiom,`error-condition || break` is more common in this situation.

tamsky:
Perhaps missing from my first comment was mentioning that the use of `break` vs `continue` here is also confusing.

If the intent is to iterate over all files that result from glob expansion, then `continue` seems appropriate...

Nuru:
@tamsky OK. See #420

PR #420, opened 2019-03-23
[direnv]  Support multiple arguments to `use_envrc`

Nuru:
## what
Following up on #418, match function execution with description, specifically: allow multiple arguments to `use_envrc`

## why
Convenience of multiple arguments vs multiple single-argument calls. 

## references
https://github.com/direnv/direnv/issues/480
#418


osterman:
Nice enhancement!

PR #421, opened 2019-03-24
Eliminate non-local effects of shopt -s nullglob

Nuru:
## what
Eliminate cases where turning on the `bash` shell's `nullglob` option lingers past its intended scope.

## why
The `bash` shell setting `shopt -s nullglob` changes the way "globs" are handled when they do not match any files. It had been being set in the prompt hook, meaning it was set for every command running in Geodesic. This caused weird behavior, like this:
```
 ⨠ ls
a-file	b-file
 ⨠ ls a*
a-file
 ⨠ ls b*
b-file
 ⨠ ls c*
a-file	b-file
```
It could likewise cause other commands and shell scripts to fail or behave unexpectedly. 

## references
[Why is nullglob not default?](https://unix.stackexchange.com/a/204944)
[Test whether a glob has any matches in bash](https://stackoverflow.com/a/34195247)

### DIFF
--- a/rootfs/etc/profile.d/terraform.sh
+++ b/rootfs/etc/profile.d/terraform.sh
@@ -1,7 +1,7 @@
 PROMPT_HOOKS+=("terraform_prompt")
 function terraform_prompt() {
 	# Test if there are any files with names ending in ".tf"
-	if compgen -G '*.tf' > /dev/null; then
+	if compgen -G '*.tf' >/dev/null; then
 		if [ ! -d ".terraform" ]; then
 			echo -e "-> Run '$(green init-terraform)' to use this project"
 		fi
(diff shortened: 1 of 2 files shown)


rootfs/etc/profile.d/terraform.sh
```diff
@@ -1,8 +1,7 @@
 PROMPT_HOOKS+=("terraform_prompt")
 function terraform_prompt() {
-	shopt -s nullglob
-	TF_FILES=(*.tf)
-	if [ ! -z "${TF_FILES}" ]; then
+	# Test if there are any files with names ending in ".tf"
+	if compgen -G '*.tf' >/dev/null; then
```
tamsky: a test for `*.tf.json` files would also seem salient here

- https://github.com/hashicorp/terraform/blob/21ba14ba1c93ca6ff087051c511601b355f9b47a/website/docs/configuration/syntax-json.html.md

PR #422, opened 2019-03-25
Enable run-time customization

Nuru:
## what
In addition to some small cleanups and additions, provide a capability for users to customize Geodesic at runtime.

## why
Because people vary, their computers vary, what they are trying to accomplish varies, and while we love to standardize, we know there is no such things as one best way for everyone.

### Customizing Geodesic

Users can place bash shell scripts on their host computer, to be read either at the start of `bash` profile script processing or at the end of it. Users can also choose whether to have a single `bash` history file for all containers or to have separate history files.

#### Root directory for configuration
All configuration files are stored under `GEODESIC_DOT_DIR`, which defaults to `/localhost/.geodesic`. At this time, `/localhost` is mapped to the host `$HOME` directory and this cannot be configured yet, so all configuration files must be under `$HOME`, but within that limitation they can be placed anywhere. So if you set `GEODESIC_DOT_DIR` to `/localhost/work/config/geodesic`, then files would go in `~/work/config/geodesic/` and below on your Docker host machine.

#### Resources
There are currently 3 Resources used for configuration:
- Preferences, which are shell scripts loaded very early in the launch of the Geodesic shell. 
- Overrides, which are shell scripts loaded very late in the launch of the Geodesic shell.
- `bash` history files, which store `bash` command line history.

Both preferences and overrides can be either a single file, named `preferences` and `overrides` respectively, or can be a collection of files in directories named `preferences.d` and `overrides.d`. If they are directories, all the visible files in the directories will be sourced, except for hidden files and files with names matching the `GEODESIC_AUTO_LOAD_EXCLUSIONS` regex, which defaults to `(~|.bak|.log|.old|.orig|.disabled)$`. `bash` history is always stored in a single file named `history`.

#### Configuration by file placement
Resources can be in several places, and will be loaded from most general to most specific, according to the name of the docker container image. 

- The most general resources are the ones directly in `GEODESIC_DOT_DIR`. These are applied first. To keep the top-level directory less cluttered and to avoid name clashes, you can put them in a subdirectory named `defaults`. If that subdirectory exists, then `GEODESIC_DOT_DIR` itself is not searched.
- The `DOCKER_IMAGE` name is then parsed. Everything before the final `/` is considered the "company" name and everything after is, following the Cloudposse reference architecture, referred to as the "stage" name. So for the `DOCKER_IMAGE` name `cloudposse/geodesic`, the company name is `cloudposse` and the stage name is `geodesic`
- The next place searched for resources is the directory with the same name as the company name: `$GEODESIC_DOT_DIR/$company`, or in our example, `~/.geodesic/cloudposse`. Here you can put resources that are common to all the containers for that company. Here again you can use a `defaults` subdirectory instead of (but not in addition to) putting resources directly here.
- The next place searched for resources is the directory with the same name as the "stage", which is generally the name of the project. In our example, that would be `~/.geodesic/geodesic`. Resources here would apply to all containers with the same base name, perhaps various forks of the same project.
- The final place searched is the directory with the full name of the Docker image: `$GEODESIC_DOT_DIR/$DOCKER_IMAGE`, i.e. `~/.geodesic/cloudposse/geodesic`. Files here are the most specific to this container. 

By loading them in this order, you can put your defaults at one level and then override/customize them at another, minimizing the amount of duplication needed to customize a wide range of containers. 

#### Usage details
Preferences and Overrides are loaded in the order specified above and all that are found are loaded. For history files, only the last one found is used. To start keeping separate history, just create an empty history file in the appropriate place. 

While Preferences and Override files themselves must be `bash` scripts and will be directly loaded into the top-level Geodesic shell, they can of course call other programs. You can even use them to pull configuration out of other places.

Symbolic links must be relative if you want them to work both inside Geodesic and outside of it. Symbolic links that reference directories that are not below `$HOME` on the host will not work.

In general, you should put most of your customization in the Preferences files. Geodesic (usually) takes care to respect and adapt to preferences set before it starts adding on top of them. The primary use for overrides is if you need the results of the initialization process as inputs to your configuration, or if you need to undo something Geodesic does not yet provide a configuration option for not doing in the first place. 

# Warning
One of the key benefits of Geodesic is that it provides a consistent environment for all users regardless of their local machine. It eliminates the "it works on my machine" excuse. While these new customization options can be great productivity enhancements as well as provide the opportunity to install new features to try them out before committing to installing them permanently, they can also create the kind of divergence in environments that brings back the "it works on my machine" problem. 

Therefore, we have included an option to disable the customization files: the preferences, the overrides, and the docker environment files. Simply set and export `GEODESIC_CUSTOMIZATION_DISABLED` to any value other than "false" 

### Other goodies
There are a few other goodies in this release.
- [`fzf`](https://github.com/junegunn/fzf) colors can now be configured via the new configuration methods described above. You can, in your preferences, set whatever color you want by setting `FZF_DEFAULT_OPTS` directly, or you can set `FZF_COLORS` to one of several options:
  - `solar_24` which only works on [terminals that support 24 bit color](https://gist.github.com/XVilka/8346728).
  - `solar_light` and `solar_dark`, which are recommended by the `fzf` author.
  - `mild`, which is the default, and is minimally showy
  - `dark`, `light`, `16`, or `bw`, which are the built-in themes

- We finally added `emacs` to the toolbox. It will load your `~/.emacs` configuration by default, but your `.emacs.d` directory will not be mapped. 
- In the `emacs` spirit, we have liberated `C-s`, a.k.a. Ctrl-S, so it can be used on the command line to forward-search through history. Previously it was mapped to `xoff` which stopped terminal output until you hit Ctrl-Q. 
- Since you can now create command aliases, we now install command line completions for aliases of commands that have completions. Thanks to [kopischke](https://superuser.com/users/101110/kopischke) at [superuser](https://superuser.com/a/437508).
- We disabled by default something we only installed a few months ago, our very own [`tfenv`](https://github.com/cloudposse/tfenv). It makes your shell environment variables available to [`terraform`](https://www.terraform.io/docs/cli-index.html) by copying them over to Terraform-specific environment variable, which we still kind of like, but we do not need to do it on a global basis anymore. Now you can install and uninstall those extra environment variables only when you need them, using our [`use tfenv`](https://github.com/cloudposse/geodesic/blob/65b76a9958a558a1241d0a766ceabd4f79832a75/rootfs/etc/direnv/rc.d/tfenv) extension to the [`direnv`](https://github.com/direnv/direnv) library. If you have been using `use tfenv` you will probably now notice that `direnv` is exporting a ton more variables than before. It's not that extra variables are newly available, it's that they used to be available everywhere so `direnv` did not need to load them. If you have not started using `direnv`, this would be a good time to do it. If you just have to have the old behavior back, you can set `export TFENV_LOAD_GLOBALLY=true` in your preferences.

### DIFF
--- a/rootfs/etc/profile.d/_geodesic-dotfiles.sh
+++ b/rootfs/etc/profile.d/_geodesic-dotfiles.sh
@@ -69,7 +69,7 @@ function _search_geodesic_dirs() {
 function _expand_dir_or_file() {
 	local -n expand_list=$1
 	local dir=${3-${PWD}}
-	local default_exclusion_pattern="(~|.old|.orig|.disabled)$"
+	local default_exclusion_pattern="(~|.bak|.log|.old|.orig|.disabled)$"
 	local exclude="${GEODESIC_AUTO_LOAD_EXCLUSIONS:-$default_exclusion_pattern}"
 
 	[[ -n $_GEODESIC_TRACE_CUSTOMIZATION ]] && echo trace: looking for resources of type "$resource" in "$dir"
--- a/rootfs/etc/profile.d/_geodesic-dotfiles.sh
+++ b/rootfs/etc/profile.d/_geodesic-dotfiles.sh
@@ -26,10 +26,10 @@ function _search_geodesic_dirs() {
 	local resource=$2
 	local base="${GEODESIC_DOT_DIR}"
 
-	[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: looking for resources of type "$resource"
+	[[ -n $_GEODESIC_TRACE_CUSTOMIZATION ]] && echo trace: looking for resources of type "$resource"
 
 	if [[ ! -d $base ]]; then
-		[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: "$base" is not a directory, giving up the search for "$resource"
+		[[ -n $_GEODESIC_TRACE_CUSTOMIZATION ]] && echo trace: "$base" is not a directory, giving up the search for "$resource"
 		return
 	fi
 
@@ -72,13 +72,13 @@ function _expand_dir_or_file() {
 	local default_exclusion_pattern="(~|.old|.orig|.disabled)$"
 	local exclude="${GEODESIC_AUTO_LOAD_EXCLUSIONS:-$default_exclusion_pattern}"
 
-	[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: looking for resources of type "$resource" in "$dir"
+	[[ -n $_GEODESIC_TRACE_CUSTOMIZATION ]] && echo trace: looking for resources of type "$resource" in "$dir"
 
 	for item in "${dir}/$2" "${dir}/${2}.d"/*; do
 		if [[ -f $item ]]; then
 			[[ $item =~ $exclude ]] && continue
 			expand_list+=($item)
-			[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: found "$item"
+			[[ -n $_GEODESIC_TRACE_CUSTOMIZATION ]] && echo trace: found "$item"
 		fi
 	done
 }
--- a/rootfs/etc/profile.d/_geodesic-dotfiles.sh
+++ b/rootfs/etc/profile.d/_geodesic-dotfiles.sh
@@ -1,0 +1,86 @@
+# bash funnctions to import
+
+# Given the name of a resource, _search_geodesic_dirs assembles an array of matching, Geodesic-specific resources,
+# in order from most general to most specific. They should be applied in order with the later ones overriding
+# the earlier ones.
+#
+# Every resource, for example "preferences", can be either a file or a directory.
+# * If it is a file, it is loaded directly
+# * If it is a directory, all the non-hidden files in that directory are loaded in glob sort order
+#
+# Several directories are searched for resources, in this order:
+# * $base/defaults/ ($base itself defaults to /localhost/.geodesic, can be set via GEODESIC_DOT_DIR)
+# * $base/ if and only if there is no $base/defaults/ directory
+# * $base/$(dirname $DOCKER_IMAGE)/defaults/
+# * $base/$(dirname $DOCKER_IMAGE)/ if and only if there is no $base/$(dirname $DOCKER_IMAGE)/defaults/ directory
+# * $base/$(base $DOCKER_IMAGE)/
+# * $base/$DOCKER_IMAGE/
+#
+# Usage: _search_geodesic_dirs array-ref resource-name
+# The first argument to _search_geodesic_dirs is the name of a variable. That variable will be treated as an array
+# and all the files found will be added in the order described above. The second argument is the name of the resource
+# See example usages in _preferences.sh
+
+
+function _search_geodesic_dirs() {
+	local -n search_list=$1
+	local resource=$2
+	local base="${GEODESIC_DOT_DIR}"
+
+	[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: looking for resources of type "$resource"
+
+	if [[ ! -d $base ]]; then
+		[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: "$base" is not a directory, giving up the search for "$resource"
+		return
+	fi
+
+	if [[ -d $base/defaults ]]; then
+		_expand_dir_or_file search_list "${resource}" "${base}/defaults"
+	else
+		_expand_dir_or_file search_list "${resource}" "${base}"
+	fi
+
+	local company=$(dirname "${DOCKER_IMAGE}")
+	local stage=$(basename "${DOCKER_IMAGE}")
+
+	if [[ -n $company && -d $base/$company ]]; then
+		if [[ -d $base/$company/defaults ]]; then
+			_expand_dir_or_file search_list "${resource}" "${base}/${company}/defaults"
+		else
+			_expand_dir_or_file search_list "${resource}" "${base}/${company}"
+		fi
+	fi
+
+	if [[ -n $stage && ($stage != $company) && -d $base/$stage ]]; then
+		_expand_dir_or_file search_list "${resource}" "${base}/${stage}"
+	fi
+
+	if [[ -n $DOCKER_IMAGE && -d $base/$DOCKER_IMAGE ]]; then
+		_expand_dir_or_file search_list "${resource}" "${base}/${DOCKER_IMAGE}"
+	fi
+}
+
+
+# _expand_dir_or_file is a helper function
+#
+# Usage: _expand_dir_or_file array-ref resource-name [base-directory]
+# Look for either a file named resource-name and a directory named resource-name.d
+# * If there is a file, add that file to the array-ref
+# * If there is a directory, add all the non-hidden files, except the ones matching
+#   the exclusion pattern in GEODESIC_AUTO_LOAD_EXCLUSIONS, in the directory to the array-ref
+function _expand_dir_or_file() {
+	local -n expand_list=$1
+	local dir=${3-${PWD}}
+	local default_exclusion_pattern="(~|.old|.orig|.disabled)$"
+	local exclude="${GEODESIC_AUTO_LOAD_EXCLUSIONS:-$default_exclusion_pattern}"
+
+	[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: looking for resources of type "$resource" in "$dir"
+
+	for item in "${dir}/$2" "${dir}/${2}.d"/*; do
+		if [[ -f $item ]]; then
+			[[ $item =~ $exclude ]] && continue
+			expand_list+=($item)
+			[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: found "$item"
+		fi
+	done
+}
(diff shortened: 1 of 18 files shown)


rootfs/etc/profile.d/direnv.sh
```diff
@@ -1,5 +1,4 @@
-PROMPT_HOOKS+=("direnv_prompt")
+# Install `direnv` via PROMPT_COMMAND hook
+#
 
-function direnv_prompt() {
-	eval "$(direnv hook bash)"
-}
+eval "$(direnv hook bash)"
```
osterman: heh, thought I saw it had to be part of the `PROMPT_COMMAND` (which is what our `PROMPT_HOOKS`) manages, but now when I look at the instructions I don't find that and see what you did as the correct way.

rootfs/etc/profile.d/tfenv.sh
```diff
@@ -1,3 +1,7 @@
 # Automatically export the current environment to `TF_VAR`
 # Use a regex defined in the `TFENV_WHITELIST` and `TFENV_BLACKLIST` environment variables to include and exclude variables
-source <(tfenv)
+#
+[[ ${LOAD_GLOBAL_TFENV:-false} == "true" ]] &&
```
osterman:
```suggestion
[[ ${LOAD_GLOBAL_TFENV:-false} == "true" ]] &&
```
```suggestion
[[ ${TFENV_LOAD_GLOBAL:-false} == "true" ]] &&
```

osterman: Use a proper `if` statment rather than chaining all the commands together =)

rootfs/etc/profile.d/Ω_overrides.sh
```diff
@@ -0,0 +1,10 @@
+## Load user's custom overrides
+OVERRIDE_LIST=()
+_search_geodesic_dirs OVERRIDE_LIST overrides
+for file in "${OVERRIDE_LIST[@]}"; do
+	[[ -n $GEODESIC_CUSTOM_TRACE ]] && echo trace: loading override file "$file"
```
osterman: `GEODESIC_DEBUG_LEVEL` or `GEODESIC_DEBUG` seem more inline to me to flags used by other software. What do you think?

rootfs/etc/profile.d/_geodesic-dotfiles.sh
```diff
@@ -0,0 +1,84 @@
...
+#
+# Usage: _search_geodesic_dirs array-ref resource-name
+# The first argument to _search_geodesic_dirs is the name of a variable. That variable will be treated as an array
+# and all the files found will be added in the order described above. The second argument is the name of the resource
+# See example usages in _preferences.sh
+
+function _search_geodesic_dirs() {
+	local -n search_list=$1
+	local resource=$2
+	local base="${GEODESIC_DOT_DIR}"
```
osterman: @aknysh @tamsky @goruha any thoughts on the name of this env?

- `GEODESIC_LOCAL_CONF`
- `GEODESIC_LOCAL_DIR`
- `GEODESIC_CONF_DIR` (probably confusing with `conf/`)
- `GEODESIC_DOT_DIR`
- `GEODESIC_DOT_CONF`
- `GEODESIC_......???`

rootfs/etc/profile.d/prompt.sh
```diff
@@ -5,7 +5,23 @@ shopt -s checkwinsize
 # Cache the current screen size
 export SCREEN_SIZE="${LINES}x${COLUMNS}"
 
-export PROMPT_COMMAND=prompter
+export PROMPT_COMMAND
+function _install_prompter() {
```
osterman: Can you add a note what this is doing in "plain english" because not sure I follow what it does/is needed.

rootfs/etc/profile.d/stty.sh
```diff
@@ -0,0 +1,8 @@
+# By default Ctrl-S (a.k.a C-s or ^S) stops terminal output.
+#  However, in bash, it is  mapped to forward-search (through command history),
```
osterman:
```suggestion
# However, in bash, it is  mapped to forward-search (through command history),
```

rootfs/etc/profile.d/Ω_finalize.sh
```diff
@@ -0,0 +1,17 @@
+## Perform any clean-up or post-setup actions
+
+# Set up command completion for command aliases
+
+_install_alias_completion
+
+# remove duplicates from PROMPT_COMMAND
```
osterman: Help me understand... don't know why we would have duplicates

rootfs/etc/profile.d/Ω_overrides.sh
```diff
@@ -0,0 +1,10 @@
+## Load user's custom overrides
+OVERRIDE_LIST=()
```
osterman:
```suggestion
SOURCE_OVERRIDE_LIST=()
```

osterman: Also, "override" seems to imply we need to override something. I think it could just be `SOURCE_LIST` since that is what it is. It might be adding or overriding functionality.

rootfs/etc/profile.d/Ω_overrides.sh
Nuru: The idea I had was to expose one variable that indicate a list of components to trace rather than set a debug level. See [`GEODESIC_TRACE`](https://github.com/cloudposse/geodesic/pull/422/files#diff-70d513de358e16629e7bffecc19a9612R1). `GEODESIC_CUSTOM_TRACE` is an internal variable that turns on tracing for the customization component.

rootfs/etc/profile.d/_geodesic-dotfiles.sh
Nuru: It's extra confusing already because it refers to a directory inside Geodesic ***and*** a directory on the Docker host computer.

rootfs/etc/profile.d/Ω_finalize.sh
Nuru: We typically run at Shell Level 3: We start up, spawn a subshell with `assume-role`, then start another subshell with `chamber exec`. Every time we start these subshells, we rerun all the initialization, but `PROMPT_COMMAND` is inherited, so poorly behaved plugins (specifically `kube-ps1`) keep adding themselves to the `PROMPT_COMMAND` and run multiple times for every prompt. Or they would, if we did not do something about it.

rootfs/etc/profile.d/Ω_overrides.sh
Nuru: The variable will contain the list of "override" files to source. It is a parallel construction to `PREFERENCES_LIST`. Plus it is internal and documented. I don't want to change that name.

osterman: image


Where does the `PREFERENCES_LIST` convention come form?

rootfs/etc/profile.d/_geodesic-dotfiles.sh
osterman: hence `GEODESIC_LOCAL_*` something might be less confusing?

rootfs/etc/profile.d/Ω_overrides.sh
osterman: Yea, I see - that makes sense. Is trace a keyword you're borrowing from some other language convention? (E.g. Java)

rootfs/etc/profile.d/prompt.sh
Nuru: Done. See https://github.com/cloudposse/geodesic/pull/422/commits/c55f06913a95748472c6e0d50cc5c37339a7b544

rootfs/etc/profile.d/Ω_overrides.sh
Nuru: I renamed it to `_GEODESIC_TRACE_CUSTOMIZATION`

rootfs/etc/profile.d/direnv.sh
Nuru: It does need to be part of the `PROMPT_COMMAND`, that what the eval does, it install it in the `PROMPT_COMMAND` string.

rootfs/etc/profile.d/_geodesic-dotfiles.sh
goruha: `GEODESIC_HOST_DIR` ?

rootfs/etc/profile.d/Ω_overrides.sh
Nuru: I made changes. See 
https://github.com/cloudposse/geodesic/pull/422/commits/dfadc0ed0a57f3114f119a917776b6efe00a1907

rootfs/etc/profile.d/_geodesic-dotfiles.sh
Nuru: Prior to this PR, Geodesic stored bash history files in `~/.geodesic`, and would automatically add `~/.geodesic/env` as a Docker `--env-file` to `docker run`. There was no option to change this.  `GEODESIC_DOT_DIR` is just a new option so you can store these things in some directory other than `~/.geodesic`. It says to use this configured directory instead of the Geodesic dot dir. We could take this variable out altogether and it would not change anything. It's just a new option.

rootfs/templates/wrapper
```diff
@@ -85,10 +81,19 @@ function use() {
 		fi
 	fi
 
-	# allow users to override value of GEODESIC_DEFAULT_ENV_FILE
-	local geodesic_default_env_file=${GEODESIC_DEFAULT_ENV_FILE:-~/.geodesic/env}
-	if [ -f "${geodesic_default_env_file}" ]; then
-		DOCKER_ARGS+=(--env-file=${geodesic_default_env_file})
+	if [[ ${GEODESIC_CUSTOMIZATION_DISABLED-false} == false ]]; then
```
osterman:
```suggestion
	if [[ ${GEODESIC_CUSTOMIZATION_DISABLED:-false} == false ]]; then
```

PR #425, opened 2019-03-25
[teleport] Fix permissions on teleport daemon

Nuru:
## what
Set the correct permissions on the teleport binary installed on the node instances.

## why
Incorrect permissions prohibited non-root users from accomplishing authorized tasks such as `scp`.


Nuru:
[teleport] Fix permissions on teleport daemon #425

Issue #417, opened 2019-03-21
Geodesic module displays broken character


osulli:
## what

- Broken character in CLI.

## why

- Character should be escaped across all platforms.

## Versions

```
FROM cloudposse/terraform-root-modules:0.58.1 as terraform-root-modules

FROM cloudposse/helmfiles:0.22.0 as helmfiles

FROM cloudposse/geodesic:0.77.2
```

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.10
Release:	18.10
Codename:	cosmic
```

## Screenshot

![image](https://user-images.githubusercontent.com/46930728/54748455-99a8d880-4bc9-11e9-80df-3cad4aa429e6.png)

```
 ✓   (healthera-sandbox-iac) aws ⨠ 
```

osterman:
@osulli can you explain why the markdown looks good with the `⨠` and the screenshot looks bad?

osterman:
basically it comes down to us using an unsupported UTF8 character for your terminal. we can update our default, since you're not the first person to have the problem.

osulli:
That's interesting. I didn't notice it working in the markdown. 
I'm on my phone now and it looks ok, I wonder if it just doesn't work on ubuntu entirely.

osulli:
From my Ubuntu 18 laptop:
![image](https://user-images.githubusercontent.com/46930728/55175887-10b91080-5178-11e9-8612-202aa3cb74ba.png)

It's clearly bigger than the terminal or Geodesic and clearly just a 'missing character' from ubuntu.

osulli:
![image](https://user-images.githubusercontent.com/46930728/55175940-23cbe080-5178-11e9-9ee9-38ee34fda8f7.png)

tamsky:
> 'missing character' from ubuntu.
To be clear, this is not an ubuntu-specific issue.
It's a missing glyph in the specific font or fonts being used in terminals or browsers.

Viewing this github comment in my browser, it's visible:
https://github.com/cloudposse/geodesic/issues/417#issuecomment-475809612

![image](https://user-images.githubusercontent.com/96978/55178481-b5b8f680-5142-11e9-86df-45448c1cb797.png)

tamsky:
@osulli 
what terminal program is in the screenshot?
are you able to determine what font your terminal is using?
have you tried any other fonts?

Nuru:
[...]

PR #429, opened 2019-04-01
[prompt], [fzf] Slightly better support for Unicode-challenged terminals

Nuru:
## what
- Add option to set `GEODISIC_PROMPT_GLYPHS` to set (or remove) glyphs at end of command line prompt
- Set `fzf` to use plain ASCII when `PROMPT_STYLE` is set to `plain`.

## why
Partly address issues such as #417

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,11 @@ install:
 - make init
 - make travis/docker-login
 - make packages/reinstall/shfmt
-- make bash/fmt/check
 
 script:
 - make deps
 - make lint
+- make bash/fmt/check
 - make docker/build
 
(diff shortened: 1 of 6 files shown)


Nuru:
@osterman Are you going to fix the `lint` problem?

Nuru:
@osterman `.travis.yaml` explicitly calls `make bash/fmt/check`, so there is no need for `make lint` to try to do that as well, so I just took it out. I think the `bash/fmt` recipes should reference $(CURDIR) rather than $(PWD) but I'm not sure enough about how the whole build system works to make the change right now.

PR #434, opened 2019-04-02
Fix aws vault and fix default profile

goruha:
## What
* Fix `aws-vault` and `AWS_DEFAULT_PROFILE`

## Why
* To fix `aws-config-setup`


Nuru:
This is not a fix, and I will incorporate this change in another PR.

PR #435, opened 2019-04-03
[aws-vault] Fix startup hang and other misc problems

Nuru:
## what
- Fix long delay at startup when opening TCP SYN packet is not rejected
- Fix use of `aws-vault` server when Docker is running on AWS and has a default route to the instance meta-data server
- Create `GEODESIC_CONFIG_HOME` directory if it does not exist
- Remove `sudo` from install recipe
- Add back missing bash completion script for `fzf`
- Do not create symbolic link for `fzf` bash completion if the file does not exist
- Fix trace message for config file search
- Remove GEODESIC_PATH (dead code)

## why
These are bugs that typically only show up on first or early use.

### DIFF
--- a/rootfs/etc/profile.d/aws-vault.sh
+++ b/rootfs/etc/profile.d/aws-vault.sh
@@ -11,7 +11,7 @@ function _validate_aws_vault_server() {
 	if [[ $instance == "aws-vault" ]]; then
 		_assume_active_aws_role
 	elif (($curl_exit_code == 0)); then
-		echo $(green force-starting aws-vault server because real AWS meta-data server is reachable)
+		echo "* $(green force-starting aws-vault server because real AWS meta-data server is reachable)"
 		_force_start_aws_vault_server
 	elif (($curl_exit_code == 7)) || (($curl_exit_code == 28)); then
 		echo "* $(green assume-role) will start EC2 metadata service at $(green http://169.254.169.254/latest)"
@@ -141,7 +141,7 @@ if [ "${AWS_VAULT_ENABLED:-true}" == "true" ]; then
 				# this function returns, regardless of how it returns (e.g. in case of errors).
 				trap 'export AWS_VAULT="$aws_vault" && export AWS_VAULT_SERVER_ENABLED="$aws_vault_server_enabled"' RETURN
 				unset AWS_VAULT
-				AWS_VAULT_SERVER_ENABLED=false
+				AWS_VAULT_SERVER_ENABLED="false: server serving other role"
 			else
 				echo "Type '$(green exit)' before attempting to assume another role"
 				return 1
--- a/rootfs/etc/profile.d/aws-vault.sh
+++ b/rootfs/etc/profile.d/aws-vault.sh
@@ -3,13 +3,10 @@
 function _validate_aws_vault_server() {
 	[[ ${AWS_VAULT_SERVER_ENABLED:-true} == "true" ]] || return 0
 
-	# For some reason, the obvious way did not work, and curl_exit_code was always 0
-	# local instance=$(curl -m 2 --connect-timeout 0.3 -sS -f http://169.254.169.254/latest/meta-data/instance-id/)
-	# local curl_exit_code=$? # was always 0
-
-	curl -m 2 --connect-timeout 0.3 -sS -f http://169.254.169.254/latest/meta-data/instance-id/ >/tmp/instance
-	local curl_exit_code=$?
-	local instance=$(cat /tmp/instance)
+	local instance
+	local curl_exit_code
+	instance=$(curl -m 2 --connect-timeout 0.3 -s -f http://169.254.169.254/latest/meta-data/instance-id/)
+	curl_exit_code=$?
 
 	if [[ $instance == "aws-vault" ]]; then
 		_assume_active_aws_role
--- a/rootfs/etc/profile.d/aws-vault.sh
+++ b/rootfs/etc/profile.d/aws-vault.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-
 function _validate_aws_vault_server() {
 	[[ ${AWS_VAULT_SERVER_ENABLED:-true} == "true" ]] || return 0
 
@@ -8,7 +7,7 @@ function _validate_aws_vault_server() {
 	# local instance=$(curl -m 2 --connect-timeout 0.3 -sS -f http://169.254.169.254/latest/meta-data/instance-id/)
 	# local curl_exit_code=$? # was always 0
 
-	curl -m 2 --connect-timeout 0.3 -sS -f http://169.254.169.254/latest/meta-data/instance-id/ > /tmp/instance
+	curl -m 2 --connect-timeout 0.3 -sS -f http://169.254.169.254/latest/meta-data/instance-id/ >/tmp/instance
 	local curl_exit_code=$?
 	local instance=$(cat /tmp/instance)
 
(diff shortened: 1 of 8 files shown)


rootfs/etc/profile.d/aws-vault.sh
```diff
@@ -1,30 +1,67 @@
...
+			echo "* $(green Attaching to exising aws-vault session and assuming role) $(cyan ${aws_vault})"
+			export AWS_VAULT="$aws_vault"
+			export ASSUME_ROLE=${AWS_VAULT}
+			export AWS_DEFAULT_REGION="${AWS_DEFAULT_REGION-${AWS_REGION}}"
 		fi
 	else
-		AWS_DEFAULT_PROFILE=$aws_default_profile
+		unset TF_VAR_aws_assume_role_arn
+		AWS_DEFAULT_PROFILE=${aws_default_profile}
+		AWS_VAULT_SERVER_ENABLED="get-caller-identity failed"
```
osterman:
```suggestion
		AWS_VAULT_SERVER_ENABLED="get-caller-identity failed"
```
```suggestion
		AWS_VAULT_SERVER_ENABLED="get-caller-identity-failed"
```

rootfs/usr/local/bin/aws-config-setup
```diff
@@ -59,7 +59,7 @@ crudini --set "${AWS_CONFIG_FILE}" "profile ${AWS_DEFAULT_PROFILE}" source_profi
 read -p "Setup AWS Credentials (aws-vault)? [y/n] " SETUP_VAULT
 
 if [ "${SETUP_VAULT}" == "y" ]; then
-	aws-vault add "${AWS_SOURCE_PROFILE}"
+	aws-vault --backend=file add "${AWS_SOURCE_PROFILE}"
```
osterman: we have `AWS_VAULT_BACKEND` defined so this should not be needed or at the very least is inconsistent.

Nuru: Recent users seem to have had issues with their `aws-vault` configs being written to the wrong place. Since Geodesic only works with `backend=file` I want to eliminate that failure mode. Because it is where keys get stored, it is particularly painful to repeat if it goes wrong the first time.

rootfs/etc/profile.d/aws-vault.sh
Nuru: Replacing the space with a dash seems confusing to me. We should not be having trouble having a space in the value, but if you want to eliminate the space, we should use some other character that is not part of `get-caller-identity`.

rootfs/usr/local/bin/aws-config-setup
osterman: but this only "fixes" one place in one script. It does not "fix" `assume-role` in `aws-vault.sh` and does not ensure that `aws-vault` run directly will work. Having the same setting set in (2) places is not desirable and breaks functionality with a supported env. Let's reduce the scope of changes.

rootfs/etc/profile.d/aws-vault.sh
osterman: What is confusing to me is we went from it being a boolean (`true` or `false`) to something overloaded. The fact that we actually care what the value is and then have it freeform is what I don't like b/c it's overloading the variable. So changing it to a key, makes it programatic and ensures we set a precedent for how others should write their functions. E.g. someone then does `AWS_VAULT_SERVER_ENABLED="ThisCamelCaseValue"` and someone else does `AWS_VAULT_SERVER_ENABLED="file_not_found"`, etc... I just want to make sure we establish consistency.

rootfs/etc/profile.d/aws-vault.sh
```diff
@@ -1,30 +1,67 @@
...
+		_assume_active_aws_role
+	elif (($curl_exit_code == 0)); then
+		echo $(green force-starting aws-vault server because real AWS meta-data server is reachable)
+		_force_start_aws_vault_server
+	elif (($curl_exit_code == 7)) || (($curl_exit_code == 28)); then
+		echo "* $(green assume-role) will start EC2 metadata service at $(green http://169.254.169.254/latest)"
+		AWS_VAULT_ARGS+=("--server")
+	else
+		echo "* $(red Unexpected status code $curl_exit_code while probing for meta-data server. Disabling aws-vault server.)"
+		export AWS_VAULT_SERVER_ENABLED="probe returned $curl_exit_code"
```
osterman: Standardize

rootfs/usr/local/bin/aws-config-setup
```diff
@@ -21,6 +21,27 @@ if [ -z "${AWS_ROOT_ACCOUNT_ID}" ]; then
 	exit 1
 fi
 
+if [ ! -d "/localhost/.awsvault/" ] || [ ! -w "/localhost/.awsvault/" ]; then
+	echo "/localhost/.awsvault/ is not writable"
```
osterman: I don't think this is what we want. If the directory does not exist, then we want `aws-vault` to create it.

rootfs/usr/local/bin/aws-config-setup
```diff
@@ -21,6 +21,27 @@ if [ -z "${AWS_ROOT_ACCOUNT_ID}" ]; then
...
+	exit 1
+fi
+
+if [ -d "/localhost/.awsvault/keys/" ] && [ ! -w "/localhost/.awsvault/" ]; then
```
osterman: We want `aws-vault` to create this.

rootfs/usr/local/bin/aws-config-setup
```diff
@@ -21,6 +21,27 @@ if [ -z "${AWS_ROOT_ACCOUNT_ID}" ]; then
...
+	exit 1
+fi
+
+if [ "${AWS_VAULT_BACKEND}" != "file" ]; then
+	echo "*"
+	echo "* WARNING: AWS_VAULT_BACKEND is \"${AWS_VAULT_BACKEND}\". Geodesic only supports AWS_VAULT_BACKEND=\"file\"."
+	echo "*"
+fi
+
+if [ ! -d "/localhost/.awsvault/" ] || [ ! -w "/localhost/.awsvault/" ]; then
```
osterman: This seems redundant with https://github.com/cloudposse/geodesic/pull/435/files#diff-fc3933c71539ab95454263b39985b8e2R24

rootfs/usr/local/bin/aws-config-setup
Nuru: `aws-vault` [does create that directory](https://github.com/cloudposse/geodesic/blob/258ca72a1f81a086d7e1bfbf3b8c44ee45123fb2/rootfs/etc/profile.d/aws-vault.sh#L56-L57). But it runs as a hook, and `aws-config-setup` runs on the command line, so there's always a chance of a disconnect.

rootfs/usr/local/bin/aws-config-setup
Nuru: Yes, but that's `aws-vault` not `aws-config-setup`

rootfs/etc/profile.d/aws-vault.sh
Nuru: We only ever test `AWS_VAULT_SERVER_ENABLED` for null/unset or `== "true"`. We never tested for `== "false"`. I thought it would be helpful when it is automatically disabled to set the value to a string that indicates why it is disabled. These are not values meant to be tested by conditionals elsewhere. If they were, I would support more consistent naming, but here I favor readability, because these values are intended for human consumption.

rootfs/usr/local/bin/aws-config-setup
osterman: The `aws-config-setup` script is what calls `aws-vault`. That is, in a cold start scenario a user will never have run `aws-vault add` before so these directories will not exist.

Nuru: Oh, I thought you meant [`aws-vault.sh`](https://github.com/cloudposse/geodesic/blob/258ca72a1f81a086d7e1bfbf3b8c44ee45123fb2/rootfs/etc/profile.d/aws-vault.sh#L56-L57) which will have run and created those directories. `aws-vault` is going to create `~/.awsvault` which is `/conf/.awsvault`, not `/localhost/.awsvault`.

osterman: Perhaps, but it does not create `.awsvault/keys`. I just tried it on a cold start. It creates `.awsvault` but not `.awsvault/keys`

PR #444, opened 2019-04-10
[direnv/codefresh] add initializer

osterman:
## what
* Add a script to initialize codefresh environment from direnv (`cf_export`)

## why
* We use `direnv` consistent across multiple platforms
* This allows us to use our `use ...` stdlib helpers in codefresh to facilitate terraform usage

### DIFF
--- a/rootfs/etc/direnv/codefresh
+++ b/rootfs/etc/direnv/codefresh
@@ -1,0 +1,16 @@
+# This is a `BASH_ENV` script to load `direnv` configuration inside of codefresh pipelines
+# https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html
+
+set -e
+
+# Prevent recursion
+unset BASH_ENV
+
+# Disable `direnv` output for scripts
+export DIRENV_LOG_FORMAT=
+
+# Allow current working directory
+direnv allow .
+
+# Process the `.envrc` and map it to `cf_export` commands
+direnv export json  | jq -r '. | to_entries[] | select (.key|test("DIRENV")|not) | "cf_export " + .key + "=" + (.value|tojson)' | bash -e


Nuru:
How is this going to impact use outside of codefresh? Is this going to require us to add `use codefresh` to all the `.envrc`? Is this going to `cf_export` everything in our interactive builds?

osterman:
@Nuru this is used as a `BASH_ENV` initializer and only used explicitly. 

Here's how we use it with [`atlantis`](https://github.com/cloudposse/geodesic/blob/master/rootfs/etc/init.d/atlantis.sh#L89).

When we use it with `codefresh`, it's also explicit. In other words, this file never gets loaded implicitly/automatically.

rootfs/etc/direnv/codefresh
```diff
@@ -0,0 +1,16 @@
...
+unset BASH_ENV
+
+# Disable `direnv` output for scripts
+export DIRENV_LOG_FORMAT=
+
+# Allow current working directory
+direnv allow .
+
+# Process the `.envrc` and map it to `cf_export` commands
+direnv export json  | jq -r '. | to_entries[] | select (.key|test("DIRENV")|not) | "cf_export " + .key + "=" + (.value|tojson)' | bash -e
```
tamsky: I thought the general pattern in `direnv` helpers was closer to:

`source <$( ... )`

osterman: @tamsky correct! however, in codefresh, the way to "source" is by calling `cf_export` on k=v pairs. So we're basically adapting `direnv` to also work seamlessly in codefresh.

tamsky: Thanks, that makes sense.

Any interest in formatting/wrapping lines and using `@sh` to properly escape the `.value` side?

(untested:)
```
direnv export json | 
    jq -r '. |
           to_entries[] | 
           select (.key|test("DIRENV")|not) |
           @sh "cf_export " + .key + "=" + \(.value|tojson)' | 
    bash -e
```

PR #445, opened 2019-04-10
Add Screen capture

osterman:
## what
* Add animated screen capture (svg) to demonstrate the interactive component of geodesic

## why
* Make it easier to see the utility of geodesic

## demo
* https://github.com/cloudposse/geodesic/blob/add-demo/README.md#screenshots


osterman:
@Nuru I think I used this: https://github.com/marionebl/svg-term

osterman:
> It's going to be hard to keep this up to date.

Story of my life. =) 

I'm of the belief something slightly out of date is better than nothing. We can only do the best we can do.

PR #85, opened 2017-11-09
Robust Bootstrapping

osterman:
## what
* Bootstrap directly from a docker container (inherited or not)
   ```
   docker run --rm cloudposse/geodesic:dev | bash -s dev
   ```
* Install the wrapper from a docker container
* Install the wrapper into `/usr/local/bin/$cluster` to support multiple installs pinned at differnet releases
* Setup all scaffolding from a docker container
   ```
   mkdir -p /tmp/test
   docker run -e CLUSTER=test.cloudposse.org \
    -e DOCKER_IMAGE=cloudposse/test.cloudposse.org \
    -e DOCKER_TAG=dev  \
        cloudposse/geodesic:dev -c new-project | tar -xv -C /tmp/test
   ```
   image
   image
   image
* Support `--env-file` so you can use the same container in multiple contexts (e.g. dev & prod)

## why
* Consolidate toolchain to run on a single tool docker
* Support multiple concurrent installs of different releases of geodesic
* No dependency on curl or storing the bootstrap script anywhere

PR #447, opened 2019-04-11
Mutliple tweaks

Nuru:
## what
- Force an update of packages when `cloudposse/packages` tag is changed
- Remove redundant `apk update`
- Map the host computer's `HOME` directory (via symbolic link) into the Geodesic container so that host absolute path names are usable inside Geodesic if they refer to files that are visible
- Enhance the shell prompt to show how many levels deep the current bash shell (`$SHLVL`) is
- Add a hook to run a script on the host when a Geodesic container exits
- Experimental feature: `role-server` (see below)
## why
- Docker layers are cached by the contents of the command line. Updates to packages via package repositories will make the cache of `apk add` invalid, but Docker will have no way of detecting that and will continue to use the cached packages layer. This is a partial workaround that ensures that at least when `cloudposse/packages` is explicitly updated in the Dockerfile, the `apk` packages will be updated.
- `apk add --update` will update the package indexes, so the earlier `apk update` is just a waste of time and space.
- Scripts, environment variables, copy & paste of file names, etc. now work between host and Geodesic when (a) absolute paths are used and (b) the files are in or below the `$HOME` directory and thus accessible via the `/localhost` mount. (However, `$HOME` itself remains different, so scripts and utilities that reference `$HOME` from inside Geodesic will continue to reference the Docker container's file system.)
- Commonly used commands such as `assume-role` and `make kops/shell` leave the user working in a subshell, but prior to this there was no indication of how deep you were, making it difficult to accurately exit subshells without exiting the container completely. Now a single-character indicator near the beginning of the line will provide this information.
- People are using the Geodesic customization hooks to do such things as alter the terminal window's appearance to indicate that it is running the Geodesic shell. These people need a reliable way of resetting the terminal's appearance once the shell exits. Unfortunately, a simple `bash` trap on`EXIT` from within Geodesic is not sufficient, so this PR adds an explicit exit hook called by the wrapper script that launches the Geodesic container. On exit, it runs `geodesic_on_exit` if that command is available. Environment variables such as `DOCKER_NAME` are available to `geodesic_on_exit` to help with labeling.
- Experimental new feature `role-server` works around a few issues with `aws-vault`.
### Shell levels
```
√ . (cpco-prod-admin) ~ ⨠ echo top level
top level
√ . (cpco-prod-admin) ~ ⨠ bash -l
√ : (cpco-prod-admin) ~ ⨠ bash -l
√ ⋮ (cpco-prod-admin) ~ ⨠ bash -l
√ 4 (cpco-prod-admin) ~ ⨠ bash -l
√ 5 (cpco-prod-admin) ~ ⨠ 
```
### `role-server`
Currently, about the first thing most people do when running a Geodesic shell is `assume-role`. That command does some fancy stuff to help you pick the right AWS IAM role to use and then hand off to `aws-vault` to set up the credentials. For various reasons, `aws-vault` is architected to set up the credentials, run a command with those credentials available, and then when the command is done it removes all access to the credentials. This fits the original use case of developers occasionally needing to run an authenticated command to tweak a setting or something, but it does not fit well with Geodesic where people are constantly in need of credentials to get anything done.  This is why `assume-role` by default starts a `bash` shell, which is a violation of the spirit of the initial `assume-vault` implementation. More importantly, `aws-vault` is simply not designed to correctly handle the situation when a credentials session expires and the command that is running is `bash`: it generally causes such problems with the `tty` that Geodesic quits entirely.  There are also other subtle issues that come into play when trying to run multiple shells or when trying to recover from other kinds of failures. 

For most people and most usages, `assume-role` will probably remain the best way to get started with work inside Geodesic. However, `role-server` provides an alternate workflow, like this:
1. Start your Geodesic container, say `prod.cpco.io`
1. Run `role-server` like you would run `assume-role`, but the behavior is different. After selecting a role (if there is more than 1 for you to choose from), you will start to see log output from the `aws-vault` command/server appear on the terminal. You still need to enter your passowrd and possibly your MFA token at the prompts, but after that you should see a line saying 
```
Local instance role server running on 127.0.0.1:9099
```
Now, just leave that running and...
1. In a new terminal window, run `prod.cpco.io` again. This does not launch a new Docker container, but rather launches a new `bash` shell in the same container.  Now, when the shell starts, instead of seeing something like
```
# Starting new prod.cpco.io session from cloudposse/prod.cpco.io:latest
# Exposing port 38880
* assume-role will start EC2 metadata service at http://169.254.169.254/latest
```
you will now see something like
```
# Attaching to existing prod.cpco.io session (86ce3e028211)
* Attaching to exising aws-vault session and assuming role cpco-prod-admin
```
Your shell is already authenticated (based on having access to your credentials available in your running Geodesic Docker container) and is a top-level shell. You can spin up more shells like this if you want to, for example one shell to run a long-running task like `kops rolling-update cluster` and another to do other work or monitor the progress of the command. Furthermore, when your aws-vault credentials expire, you work in these shells will only be momentarily interrupted. You will see an error about needing credentials, but can then go back to the window where you started `role-server`. You may be able to continue working by just entering your MFA token, or you may need to run `role-server` again. Either way, you simply re-authenticate and then continue on with your work in the other terminals. No more getting kicked out of shells and contexts. 

As an added bonus, you now have a screen on which you can see the full logs of the authentication process, including when credentials are rotated and how much longer they have until they expire. This will help when you have credential issues that you do not understand, or if you are planning on running a very long-running script. 

One quirk that remains, and you can choose to view it as a feature, is that when the `role-server` Geodesic shell quits, it will kill the container and remove it, causing all your other shells to exit unceremoniously. This is handy if you want to rebuild the container and want to make sure no others terminals are running it, but it can be a problem if you close the wrong window, so be careful.


Dockerfile
```diff
@@ -1,3 +1,6 @@
+# ARGs used in FROM must be declared before the first FROM
+ARG CLOUDPOSSE_PACKAGES_VERSION="0.86.0"
```
osterman: The reason we did not do this is b/c it then thwarts dependabot. We can increase the frequency of dependabot checks, which would be better.

rootfs/etc/profile.d/prompt.sh
```diff
@@ -91,6 +91,15 @@ function geodesic_prompt() {
 		;;
 	esac
 
+	local level_prompt
+	case $SHLVL in
```
osterman: Hah! neat idea. Have you seen this elsewhere? The `.` will be kind of odd, no?

rootfs/etc/profile.d/prompt.sh
```diff
@@ -91,6 +91,15 @@ function geodesic_prompt() {
...
+	1) level_prompt='.' ;;
+	2) level_prompt=':' ;;
+	3) level_prompt='⋮' ;; # vertical elipsis \u22ee from Mathematical Symbols
+	*) level_prompt="$SHLVL" ;;
```
osterman: Maybe? Haven't seen what it looks like.
```suggestion
	*) level_prompt="($SHLVL)" ;;
```

rootfs/templates/wrapper
```diff
@@ -46,6 +46,10 @@ function debug() {
 	fi
 }
 
+function run_exit_hooks() {
+	command -v on_geodesic_exit >/dev/null && on_geodesic_exit
```
osterman:
```suggestion
	command -v geodesic.exit >/dev/null && geodesic.exit
```
Perhaps use `geodesic.$hook`

Dockerfile
```diff
@@ -53,13 +56,16 @@ RUN echo "@cloudposse https://apk.cloudposse.com/3.9/vendor" >> /etc/apk/reposit
...
+    echo "@community https://alpine.global.ssl.fastly.net/alpine/edge/community" >> /etc/apk/repositories
 
 # Install alpine package manifest
 COPY packages.txt /etc/apk/
 
-RUN apk add --update $(grep -v '^#' /etc/apk/packages.txt) && \
+
+# Import package version for cache busting
+ARG CLOUDPOSSE_PACKAGES_VERSION
+RUN echo Using cloudposse/packages version ${CLOUDPOSSE_PACKAGES_VERSION} as packages cache key && \
```
osterman: Aha, I see... So that forces this to re-run. What if we used `ADD ... https://apk....APKINDEX` instead? I think `ADD` will invalidate the cache if the contents of the source change.

osterman: What I don't like about this hack is that there's actually nothing more than optimistic correlation between `CLOUDPOSSE_PACKAGES_VERSION` and the packages installed.

rootfs/etc/profile.d/prompt.sh
Nuru: Of course Geodesic is filled with contex-dependent prompt information, but I have never been in a situation like this where I am routinely in subshells but don't know how deep I am, so no, I've never seen this particular feature before.

rootfs/templates/wrapper
Nuru: Are you sure? This is going to need to be installed on the user's `$PATH` on their host computer.

Makefile
```diff
@@ -30,3 +30,9 @@ bash/fmt:
 
 bash/fmt/check:
 	shfmt -d $(PWD)/rootfs
+
+apk-update:
+	@echo package_repos.md5 old $$(cat package_repos.md5 || echo '<not found>')
+	@docker run --rm -e CLUSTER=galaxy $(DOCKER_IMAGE_NAME) -c \
```
osterman: galaxy?

Makefile
```diff
@@ -30,3 +30,9 @@ bash/fmt:
 
 bash/fmt/check:
 	shfmt -d $(PWD)/rootfs
+
+apk-update:
```
osterman: Not crazy about the external (somewhat circular) dependency introduced here. Would really prefer the `ADD` solution if it works since that's a one-liner.

osterman: Also, the `ADD` workaround if it works is fully self-contained and doesn't have "hooks" outside of the `Dockerfile`

rootfs/etc/profile.d/_preferences.sh
```diff
@@ -13,6 +13,13 @@ fi
 
 [[ -n $_GEODESIC_TRACE_CUSTOMIZATION ]] && echo trace: GEODESIC_CONFIG_HOME is found to be "${GEODESIC_CONFIG_HOME:-<unset>}"
 
+# If HOST_HOME is set, create a symbolic link so host pathnames (at least the ones under $HOME) work inside the shell
```
osterman: How is `HOST_HOME` different from `LOCAL_HOME` and can we possible consolidate on `LOCAL_HOME`?

rootfs/templates/wrapper
```diff
@@ -85,6 +89,7 @@ function use() {
 		if [ -n "${ENV_FILE}" ]; then
 			DOCKER_ARGS+=(--env-file ${ENV_FILE})
 		fi
+		DOCKER_ARGS+=(--env HOST_HOME=$HOME)
```
osterman: Keep in mind `HOME` is "complicated" since we work with "WSL" as well. That's why we extrapolate the `LOCAL_HOME`

Dockerfile
```diff
@@ -53,16 +53,19 @@ RUN echo "@cloudposse https://apk.cloudposse.com/3.9/vendor" >> /etc/apk/reposit
...
 RUN sed -i 's|http://dl-cdn.alpinelinux.org|https://alpine.global.ssl.fastly.net|g' /etc/apk/repositories && \
     echo "@testing https://alpine.global.ssl.fastly.net/alpine/edge/testing" >> /etc/apk/repositories && \
-    echo "@community https://alpine.global.ssl.fastly.net/alpine/edge/community" >> /etc/apk/repositories && \
-    apk update
+    echo "@community https://alpine.global.ssl.fastly.net/alpine/edge/community" >> /etc/apk/repositories
 
 # Install alpine package manifest
 COPY packages.txt /etc/apk/
+# Install repo checksum in an attempt to ensure updates bust the Docker build cache
+COPY package_repos.md5 /var/cache/apk/
```
osterman: Have you tried doing this instead:

```
ADD https://apk.cloudposse.com/3.9/vendor/x86_64/APKINDEX.tar.gz /var/tmp/APKINDEX.tar.gz
```
(or the like?)

My underestanding is `ADD` will evaluate the contents of the endpoint to determine if the layer has changed.

osterman: Also, while `COPY` is generally the best practice, it doesn't mean using `ADD` is wrong; there's a time and place for it.

rootfs/templates/wrapper
osterman: This is a command that the user can optionally have installed on their machine, right? Then a prefix of `geodesic` makes more sense to me, rather than a prefix of `on_`. 

We can maybe do `geodesic.on_exit` or something like that?

Makefile
Nuru: Using `sed` to edit the repositories file made it look to me like you strongly wanted to avoid explicitly listing URLs, and ADD will only work if you are going to fully specify the URL for each repo index file. If you want to do that, then yes, that is a good solution because it is self contained, although it will very much slow me down because on every build it will have to fetch all the indexes just to verify that they haven't changed. 

The `make` approach lets you explicitly determine when to update packages, and effectively pins the packages in between updates to the md5 fie.

Dockerfile
Nuru: You have to do that for each package repo, and it executes on every build, which slows me down a lot (okay, not a lot, but it's annoying). It also yields the opposite problem to the one we have now. Right now, we are missing package updates we want because build layers are cached. With ADD, we will get package updates as soon as they are released whether we want them or not, and without warning or notification.

Makefile
Nuru: I don't understand half of what is in the existing Makefile recipes. I just copied `galaxy` from https://github.com/cloudposse/geodesic/blob/05f4ab3001cd561a7b7b1fd5fbd931c9a8cb712f/Makefile#L23

rootfs/etc/profile.d/_preferences.sh
Nuru: I didn't look carefully at `LOCAL_HOME` before, but it is what I need and I will use it.

rootfs/templates/wrapper
Nuru: OK, but no dots. Dots have way too much significance. I like to blame DOS but it really started as early as Unix and C.

Makefile
osterman: Hrmmm I wonder if we still need `CLUSTER`; I think it's an orphaned environment variable.

Dockerfile
osterman: > With ADD, we will get package updates as soon as they are released whether we want them or not, and without warning or notification.

Agree. 

> With the explicit make of the md5 file, we get something closer to npm, bundler, bazel where the packages only change when we explicitly ask them to. This method is far from foolproof, but is at least a good compromise in that direction.

Agree. It's a step in the right direction.

Dockerfile
```diff
@@ -53,16 +53,19 @@ RUN echo "@cloudposse https://apk.cloudposse.com/3.9/vendor" >> /etc/apk/reposit
...
 
 RUN apk add --update $(grep -v '^#' /etc/apk/packages.txt) && \
     mkdir -p /etc/bash_completion.d/ /etc/profile.d/ /conf && \
     touch /conf/.gitconfig
 
+RUN [[ $(md5sum /var/cache/apk/APKINDEX.* | md5sum | colrm 33) == $(cat /var/cache/apk/package_repos.md5) ]] || echo WARNING: apk package repos mismatch 1>&2
```
osterman:
```suggestion
RUN [[ $(md5sum /var/cache/apk/APKINDEX.* | md5sum | colrm 33) == $(cat /var/cache/apk/apkindex.md5) ]] || echo WARNING: APKINDEX mismatch 1>&2
```

Makefile
```diff
@@ -30,3 +30,9 @@ bash/fmt:
...
+	@echo package_repos.md5 old $$(cat package_repos.md5 || echo '<not found>')
+	@docker run --rm -e CLUSTER=galaxy $(DOCKER_IMAGE_NAME) -c \
+	'apk update >/dev/null && md5sum /var/cache/apk/APKINDEX.* | md5sum | colrm 33' > package_repos.md5
```
osterman:
```suggestion
	'apk update >/dev/null && md5sum /var/cache/apk/APKINDEX.* | md5sum | colrm 33' > apkindex.md5
```

osterman: This really has nothing to do with `packages.txt` and more to do with the `APKINDEX`. Something like `APKINDEX.md5` or `.APKINDEX` or `.APKINDEX.md5` would be better names IMO.

Nuru: Using `APKINDEX.md5` has a high probability of interfering with, or at least being mistaken for, something used by `apk`. Since the `apk` documentation is sparse, I hate to do anything potentially confusing like that.

I'm not particularly advocating for `package_repos.md5`, but I do want something that will not be mistaken for part of the official `apk` integrity checks.

Makefile
Nuru: @osterman you added CLUSTER=galaxy. See https://github.com/cloudposse/geodesic/blob/2a97ad046a63db46ec9b34cb7b4559b518c2885a/README.md#quick-start and https://github.com/cloudposse/geodesic/pull/85

Makefile
```diff
@@ -30,3 +30,10 @@ bash/fmt:
 
 bash/fmt/check:
 	shfmt -d $(PWD)/rootfs
+
+.PHONY: geodesic_apkindex.md5
+apk-update geodesic_apkindex.md5:
+	@echo geodesic_apkindex.md5 old $$(cat geodesic_apkindex.md5 || echo '<not found>')
```
osterman:
```suggestion
	@echo $@ old $$(cat geodesic_apkindex.md5 || echo '<not found>')
```

Dockerfile
```diff
@@ -53,16 +53,21 @@ RUN echo "@cloudposse https://apk.cloudposse.com/3.9/vendor" >> /etc/apk/reposit
...
     echo "@testing https://alpine.global.ssl.fastly.net/alpine/edge/testing" >> /etc/apk/repositories && \
-    echo "@community https://alpine.global.ssl.fastly.net/alpine/edge/community" >> /etc/apk/repositories && \
-    apk update
+    echo "@community https://alpine.global.ssl.fastly.net/alpine/edge/community" >> /etc/apk/repositories
 
 # Install alpine package manifest
 COPY packages.txt /etc/apk/
+# Install repo checksum in an attempt to ensure updates bust the Docker build cache
+COPY geodesic_apkindex.md5 /var/cache/apk/
+COPY rootfs/usr/local/bin/geodesic-apkindex-md5 /tmp/
```
osterman: Why `/tmp`? I think we can keep it.

Dockerfile
```diff
@@ -53,16 +53,21 @@ RUN echo "@cloudposse https://apk.cloudposse.com/3.9/vendor" >> /etc/apk/reposit
 # Use TLS for alpine default repos
 RUN sed -i 's|http://dl-cdn.alpinelinux.org|https://alpine.global.ssl.fastly.net|g' /etc/apk/repositories && \
```
osterman: I'm not beholden to this. Do you think we're better off moving all those to `rootfs/etc/apk/repositories`?

Dockerfile
Nuru: Because active development if Geodesic mainly involves making changes to rootfs, and the full build of Geodesic takes over 10 minutes, I want to keep the copy of rootfs as the last step. So we temporarily install the script in /tmp so we can run it now. We later (at the end of the Dockerfile) install it permanently when we copy over the entire rootfs.

Dockerfile
Nuru: I think the `sed` approach is better in that it is somewhat self-documenting about why we are not using the default repositories, and it updates itself when we upgrade Alpine.

Dockerfile
osterman: I guess my point was why not just `COPY rootfs/usr/local/bin/geodesic-apkindex-md5 /usr/local/bin`; I understand why not to copy everything.

Makefile
```diff
@@ -20,7 +20,7 @@ build:
 	@make --no-print-directory docker:build
 
 install:
-	@docker run --rm -e CLUSTER=galaxy $(DOCKER_IMAGE_NAME) | bash -s $(DOCKER_TAG) || (echo "Try: sudo make install"; exit 1)
+	@docker run --rm -e $(DOCKER_IMAGE_NAME) | bash -s $(DOCKER_TAG) || (echo "Try: sudo make install"; exit 1)
```
osterman: Thanks

PR #452, opened 2019-04-17
[tsh] Initial support for teleport tsh

Nuru:
## what
Initial support for [Teleport](https://gravitational.com/teleport)
## why
Customer request plus Teleport support for Geodeisc


rootfs/templates/wrapper
```diff
@@ -150,7 +150,7 @@ function use() {
 		--publish ${GEODESIC_PORT}:${GEODESIC_PORT}
 		--name "${DOCKER_NAME}"
 		--rm
-		--env KUBERNETES_API_PORT=${GEODESIC_PORT})
+		--env GEODESIC_PORT=${GEODESIC_PORT})
```
goruha: I'm not sure about ports

goruha: @osterman

osterman: I support this. I want to generalize it so we can use it for both `kubectl` and `teleport`. The port can be overloaded,

osterman: Also, I like the consistency where the LHS==RHS env name

PR #453, opened 2019-04-18
[tsh] Use official tsh environment variables for port, proxy, & ssh login

Nuru:
## what
[tsh-login] Distinguish login proxy name from local cluster name
## why
There are 2 relevant proxy names for trusting clusters and they are used differently

### DIFF
--- a/rootfs/etc/profile.d/tsh-login.sh
+++ b/rootfs/etc/profile.d/tsh-login.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 function tsh-login() {
-	if (( $# == 0 )); then
+	if (($# == 0)); then
 		tsh login --bind-addr=:${GEODESIC_PORT} --proxy=${TELEPORT_LOGIN_PROXY:-tele.${DOCKER_IMAGE#*.}} $STAGE
 	else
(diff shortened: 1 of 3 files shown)


rootfs/etc/profile.d/tsh-login.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
 function tsh-login() {
-	tsh login --bind-addr=:${GEODESIC_PORT} --proxy=${TELEPORT_PROXY_DOMAIN_NAME:-tele.${DOCKER_IMAGE#*.}} $*
+	if (($# == 0)); then
+		tsh login --bind-addr=:${GEODESIC_PORT} --proxy=${TELEPORT_LOGIN_PROXY:-tele.${DOCKER_IMAGE#*.}} $STAGE
```
osterman: Instead, we shouold export the envs like `TELEPORT_LOGIN_BIND_ADDR` so we can call `tsh login` without arguments

osterman: https://github.com/gravitational/teleport/issues/2620#issuecomment-481756625

osterman: And use https://github.com/gravitational/teleport/blob/7e636b0b2b14bee3b4a0103cedcf8ddde9dc082a/tool/tsh/tsh.go#L188 instead of `--proxy`

PR #456, opened 2019-04-23
[kops] K8S private 

goruha:
## What
* Make k8s LB private

## Why
* To protect k8s access from the outside


goruha:
@osterman @Nuru ok - but let's merge this for back compatibility.

goruha:
@osterman @Nuru 
I created the PR to reference architecture https://github.com/cloudposse/reference-architectures/pull/26

PR #467, opened 2019-05-04
Add support for shell documentation

osterman:
## what
* Add support for `man` pages from markdown

## why
* Geodesic has insufficient documentation in part because we haven't had a good way to deliver it
* Unix has solved this: `man` pages have been there forever. The problem is that `man` (`groff`) is no fun to write. Using `pandoc` we can convert markdown to valid man pages to make it easy to contribute documentation

## references
* http://rtomayko.github.io/ronn/ronn-format.7.html (did not want to install `ruby` just to compile docs; using `pandoc` instead)
* https://pandoc.org/MANUAL.html
* https://en.wikipedia.org/wiki/Man_page
* https://babbage.cs.qc.cuny.edu/courses/cs701/Handouts/man_pages.html


rootfs/usr/local/bin/docs
```diff
@@ -0,0 +1,74 @@
...
+
+tasks:
+  update:
+    description: "Rebuild man pages and reindex"
+    steps:
+    - task: "build"
+    - task: "index"
+
+  build:
+    # Write out kubecfg
```
aknysh:
```suggestion
    # Build man pages
```

geodesic_apkindex.md5
```diff
@@ -1 +0,0 @@
-77e3cde7afac220dbf00e433fbaf4abb
```
Nuru: You should not have an empty `geodesic_apkindex.md5`. Do you maybe have `*.md5` in your global `.gitignore`?

osterman: I ran the update command but didn’t pay attention to the result. Will check what’s wrong next time.

PR #469, opened 2019-05-06
Move python requirements to alpine packages

osterman:
## what
* deprecate `awsebcli` 
* install `ansible`, `awscli` from apk
* remove pinning on python packages `boto` and `Jinja2`

## why
* Fix broken travis builds due to python deps timing out
* Speed up docker builds


osterman:
Switched to codefresh from travis and everything works. Closing this for now.

PR #472, opened 2019-05-09
[helmctl] Add missing key

osterman:
## what
* Add `parameters` section

## why
* Not sure how it got dropped


osterman:
Please share details.

This is what I see.

image

Nuru:
Somehow I got a tab instead of spaces when I inserted the change into my file. Works now.

PR #475, opened 2019-05-10
istioctl helper script

maximmi:
## what
* ctl script to help work with istio

## why
* to reuse it in nested geodesic containers instead of copying to every one


rootfs/usr/local/bin/istioctl
```diff
@@ -1,21 +1,40 @@
...
+  injection:
     description: Enable Istio Sidecar Injection for a namespace
     script: |
-      kubectl label namespace {{ get "namespace" }} istio-injection=enabled --overwrite=true
\ No newline at end of file
+      kubectl label namespace {{ get "namespace" }} istio-injection=enabled --overwrite=true
+
+  init:
+    description: Install Istio helm chart
+    script: |
```
osterman: We should clone to a temporary directory

rootfs/usr/local/bin/istioctl
```diff
@@ -1,21 +1,40 @@
...
 
   # https://istio.io/docs/setup/kubernetes/quick-start/
-  istio-injection:
+  injection:
     description: Enable Istio Sidecar Injection for a namespace
     script: |
-      kubectl label namespace {{ get "namespace" }} istio-injection=enabled --overwrite=true
\ No newline at end of file
+      kubectl label namespace {{ get "namespace" }} istio-injection=enabled --overwrite=true
+
```
osterman: We need an `install` command too, since `init` only installs the CRDs

rootfs/usr/local/bin/istioctl
```diff
@@ -1,21 +1,40 @@
...
+  init:
+    description: Install Istio helm chart
+    script: |
+      git clone --depth 1 --branch {{ get "version" }} {{ get "repo" }}
```
osterman: The clone should be shared by an `init` and `install` target

osterman: (so move `clone` to a new command)

osterman:
`istioctl` is already used by an official command. Let's rename to `istioadm`

PR #477, opened 2019-05-18
Force git to use HTTPS instead of SSH for atlantis

aknysh:
## what
* Force `git` to use HTTPS instead of SSH for `atlantis`

## why
* Even when `git@github.com:` protocol is specified for private GitHub repos, `git` will rewrite the URLs to use HTTPS protocol, for which we can use `GITHUB_TOKEN` (instead of SSH keys) to authenticate with GitHub when accessing private repos from `atlantis`

## references
* https://ricostacruz.com/til/github-always-ssh

### DIFF
--- a/rootfs/etc/init.d/atlantis.sh
+++ b/rootfs/etc/init.d/atlantis.sh
@@ -82,11 +82,17 @@ if [ "${ATLANTIS_ENABLED}" == "true" ]; then
 	# Do not export these as Terraform environment variables
 	export TFENV_BLACKLIST="^(AWS_ACCESS_KEY_ID|AWS_SECRET_ACCESS_KEY|AWS_SECURITY_TOKEN|AWS_SESSION_TOKEN|ATLANTIS_.*|GITHUB_.*)$"
 
-    # Configure Git credentials for atlantis to allow access to GitHub private repos (see rootfs/usr/local/bin/git-credential-github)
-    # https://git-scm.com/book/en/v2/Git-Tools-Credential-Storage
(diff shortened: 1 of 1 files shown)


rootfs/etc/init.d/atlantis.sh
```diff
@@ -82,11 +82,19 @@ if [ "${ATLANTIS_ENABLED}" == "true" ]; then
...
+    # Configure git credentials for atlantis to allow access to GitHub private repos
     export GITHUB_USER=${ATLANTIS_GH_USER}
     export GITHUB_TOKEN=${ATLANTIS_GH_TOKEN}
     chmod +x /usr/local/bin/git-credential-github
+
+    # Force git to use HTTPS instead of SSH. With HTTPS, git will use the `GITHUB_TOKEN` to authenticate with GitHub (with SSH it won't)
+    # https://ricostacruz.com/til/github-always-ssh
+    # https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf
+    # https://gist.github.com/Kovrinic/ea5e7123ab5c97d451804ea222ecd78a
+    git config --global url."https://github.com/".insteadOf git@github.com:
```
osterman: Quotes are not needed as in original example?

aknysh: added quotes
but they have examples with quotes and without on the same page
https://ricostacruz.com/til/github-always-ssh
https://gist.github.com/Kovrinic/ea5e7123ab5c97d451804ea222ecd78a

(maybe just `git://github` does not require quotes)

PR #478, opened 2019-05-20
Update kops, remove kops & kubernetes version from Dockerfile, add aliases

Nuru:
## what
- Update kudo from 1.11.1 to 1.12.1
- Update kubectl from 1.13.0 to 1.14.2
- Remove settings for `KUBERNETES_VERSION` and `KOPS_BASE_IMAGE`
- Minor bugfixes to `assume-role` and `role-server`
- Add new commands as aliases
## why
- Support Kuberenetes 1.12
- Support Kubernetes 1.14, bring in bug fixes
- Default settings in Dockerfile have turned out to cause problems with consistency within companies. Removed in favor of setting `KUBERNETES_VERSION` and `KOPS_BASE_IMAGE` in `/conf/kops/kops.envrc`
- Replace subshells with simple commands
#### `chudo`/ `kudo` and `chundo`/`kundo`
Named in reference to the `sudo` command and "undo" operation, these commands replace 2 previous commands that create subshells with secrets that enable `kops` to control a Kubernetes cluster.

Previously:
```
cd /conf/kops
make deps
make kops/shell
```
or
```
chamber exec kops -- bash -l
```
Now: `kudo` or `chdo kops` import those secrets into the current shell rather than creating a subshell. 

Previously: you removed the secrets from the shell by exiting the shell.
Now: `kundo` or `chundo kops` remove the secrets from the current shell.

`kudo` is shorthand to just import the `chamber` secrets in the `kops` service. `chudo` is more generic, taking a list of services, e.g. `chudo kops datadog codefresh`. Likewise, `chundo` takes a list of services, and `kundo` is just an alias for `chundo kops`.

#### `kexp`
Alias for `kops export kubecfg`. So now the standard way to enable `kubectl` (without using `teleport` or some other add-on) is:
```
assume-role
kudo
kexp
```

#### `assume_active_aws_role`
For the case where you have an active `aws-vault` server but the current shell is not using it, you can run `assume_active_aws_role` to assume the role being served by the server.  Normally this is run automatically for you when the shell starts but if you start server later, you can now run this manually.

#### `taboff` and `tabon`
By default, `bash` command completion is turned on, and typing <tab> on the command line will trigger auto-completion of the current command. This causes problems, however, if you are pasting shell commands from a file that uses tabs for indenting. In this case, you can now use `taboff` to disable this special handling of the <tab> character, paste your text with tabs, and then use `tabon` to restore tab completion.


rootfs/etc/profile.d/aws-vault.sh
```diff
@@ -14,7 +14,7 @@ function _validate_aws_vault_server() {
 
 	if [[ $instance == "aws-vault" ]]; then
 		# We successfully connected to an active aws-vault credentials server, so use it
-		_assume_active_aws_role
```
osterman: Why the change? Guessing this was done previously to be similar to https://github.com/cloudposse/geodesic/pull/478/files#diff-1ae3ba81cafb63b08cd75d2558472b27R230

Nuru: The leading underscore was intended to hide the function from the namespace of commands people would run on the command line. Turns out this command has utility on the command line (see usage notes in the PR), so I changed the name to reflect that.

rootfs/etc/profile.d/aliases.sh
```diff
@@ -4,9 +4,36 @@ alias default='kubectl --namespace=default'
...
 alias ll='ls -l'
 
+# taboff turns off tab auto-completion, which is practically required when pasting code with tabs into the terminal
+alias taboff='bind '\''set disable-completion on'\'''
+# tabon turn on tab auto-completion
+alias tabon='bind '\''set disable-completion off'\'''
+
+
+# Run chamber to put secrets for given service(s) in the environment
+function chdo() {
```
osterman: How about adding a small snippet in `docs/chdo.md` that basically explains this. I know it's tedious, but I think it will get easier once we have a few.

osterman: Atleast it would help in discovery of commands and provide feedback on how we can make our `man` pages better

PR #482, opened 2019-05-28
Update issue templates

osterman:
## what
* Update bug report template

## why
* Promote richer bug reports


Nuru:
I'm not generally happy with these templates. Let's take a step back and look for better starting points.

For a bug report template, we should have
- a 1-line description
- a short summary
- what you did (step to reproduce)
- what you expected to happen
- what actually happened
- environment/versions
- known workaround, if any
- impact of the bug (how much work is it causing/how much work would fixing this bug save)


Pull request template should include
- Any lint/format/readme steps required to make the PR acceptable
- Any other testing required
- Any other documentation or other requirements
- Description of what, if any, testing has been done 

Google around and look at existing templates. For example, here's a whole repo full: https://github.com/stevemao/github-issue-templates

PR #492, opened 2019-06-21
[helm] Update to helm 2.14.1

Nuru:
## What
* Unpin helm, update to version to `2.14.1`

## Why
* 2.14.0 does not work with `cert-manager` among other things. See https://github.com/helm/helm/issues/5750
* Pinning versions in Geodesic is counter-productive. Users should pin versions in their environments as needed.


packages.txt
```diff
@@ -28,7 +28,7 @@ gomplate@cloudposse
 goofys@cloudposse
 gosu@cloudposse
 groff
-helm@cloudposse==2.14.0-r0
+helm@cloudposse==2.14.1-r0
```
osterman:
```suggestion
helm@cloudposse
```

osterman: (this is not directed at you @Nuru, just from a previous discussion on Slack)

lets *never* rely on pinning versions in `geodesic` to fix problems of individual customers.
let's pin versions for customers in their own `geodesic` environment.
otherwise we're just playing a constant game of fixing things in one place, while breaking it in another.
pinning only works if we're surgical about it; updating pinning in `geodesic` and then updating the customer to point to it is about as stable as pinning to `master`.
and that's not stable

PR #499, opened 2019-07-12
[aliases] Do not replace existing aliases

Nuru:
## what
1. [aliases] Do not overwrite existing commands with our default aliases and functions
1. [codefresh] Add some helper functions to be used with Codefresh CI/CD pipeline
1. [tfenv] Set a default `TFENV_BLACKLIST` if the variable is not already one set
1. [docker] Tag docker images with short commit tags as well as long ones

## why
1. If a user has customized Geodesic in a way that conflicts with our defaults, let the user's customization win. Better user experience.
1. Move toward using Geodesic rather than Build Harness inside CI/CD pipelines.
1. Prevent secrets in the environment from leaking into Terraform. 
1. Short tags are easier for humans to deal with, and are more visible in the GitHub UI.


rootfs/usr/local/bin/helmfile-directory
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
```
osterman: I think this should be generalized to config-directory and take an argument which is the project (e.g Helmfile)

Nuru: Replaced with `conf-directory helmfiles`

PR #511, opened 2019-08-18
Add grafana-db tool, pin PyYAML

Nuru:
## what
1. Add `grafana-db` tool to install Grafana dashboards via Grafana API
1. Pin PyYAML to version 3.13
1. Update `cloudposse/packages`
1. Install Python 3

## why
1. Dashboards can be sourced from URLs and once added, immediately edited (vs. provisioned dashboards which must be exported and re-imported to become editable)
1. PyYAML 3.13 is the latest version that also satisfies constraints of `awscli` and `awsebcli`
1. Update `kops`, `helm`, `hemlfile` etc.
1. Current tools use Python 3 and it is a pain to install on the fly.

## other notes
I attempted to drop Python 2 and use only Python 3. That did not work because at least `google/cloud-sdk` and `crudini` still only work with Python 2. 

For the record, I tried replacing `FROM alpine:3.10.1 as python` with `FROM python:3.7.4-alpine3.10 as python` and not installing the Python packages. That seemed to be a better way to build Python stuff, since the packages can be quite behind the current release. However, that caused problems because the shell scripts that installed themselves (such as `aws_completer`) created shebang lines that referenced the `python` install where it was in the `python` image (`/usr/local/bin`) which is not where it ended up in the final Geodesic build (`/usr/bin`), so they broke. Possibly basing the final Geodisic image on the `python` image would be a better way to stay up to date on Python.

Until then, here is how I was able to build everything under Python 3.
```
## When we are ready for Python3, here is how:
## Package pruning
##   python and python-dev: replace with python3 and python3-dev
##   py-pip: only needed for python 2
##   py-virtualenv: replaced with built-in venv in Python 3
RUN sed -i 's|http://dl-cdn.alpinelinux.org|https://alpine.global.ssl.fastly.net|g' /etc/apk/repositories
RUN apk add python3 python3-dev libffi-dev gcc linux-headers musl-dev openssl-dev make

COPY requirements.txt /requirements.txt

RUN python3 -m pip install --upgrade pip setuptools wheel && \
    pip install -r /requirements.txt --install-option="--prefix=/dist" --no-build-isolation
```


Dockerfile
```diff
@@ -3,8 +3,22 @@
 #
 FROM alpine:3.10.1 as python
 
+## Sadly, not ready for Python 3 yet. Crudini and gcloud SDK not updated.
+## When we are ready, changes below ready to go
+## Package pruning
+##   python and python-dev: replace with python3 and python3-dev
+##   py-pip: only needed for python 2
+##   py-virtualenv: replaced with built-in venv in Python 3
+#RUN sed -i 's|http://dl-cdn.alpinelinux.org|https://alpine.global.ssl.fastly.net|g' /etc/apk/repositories
```
osterman: Why commented out?

Dockerfile
```diff
@@ -3,8 +3,22 @@
...
+#RUN apk add python3 python3-dev libffi-dev gcc linux-headers musl-dev openssl-dev make
+#
+#COPY requirements.txt /requirements.txt
+#
+#RUN python3 -m pip install --upgrade pip setuptools wheel && \
+#    pip install -r /requirements.txt --install-option="--prefix=/dist" --no-build-isolation
```
osterman: Why commented out?

Dockerfile
osterman: Maybe add python 3 stuff to comments of PR somewhere for your own records. I don't want to keep code commented out in the `Dockerfile` as it's just techdebt.

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+#  PROMETHEUS_GRAFANA_ROOT_URL or GRAFANA_ROOT_URL: URL to which '/api/dashboards...' will be appended
+#
+# Usage:
+#    grafana-db <URL or gnetId>
+#
+#    URL is URL to (raw) ConfigMap or ConfigMapList containing Grafana dashboards as for Prometheus Operator
+#    gnetId is ID (typically a 4 digit number) for the dashaboard on Grafana.com
+
+## TODO
+## Allow uploading a local file. Main this is determining, detecting, and parsing format of file
```
osterman:
```suggestion
## Allow uploading a local file. Main thing is determining, detecting, and parsing format of file
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+## Remove absolute times in .time
```
osterman:
```suggestion
## - Remove absolute times in .time
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+##   Bonus points: calculate the time interval in .time and replace with relative time interval of same length
+##                 round to standard relative time choice
+## Better error handling
```
osterman:
```suggestion
## - Better error handling
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+## Optional overwrite flag
```
osterman:
```suggestion
## - Optional overwrite flag
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+## Allow host and api_key on command line
```
osterman:
```suggestion
## - Allow host and api_key on command line
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+## Optional deletion of UID
```
osterman:
```suggestion
## - Optional deletion of UID
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+## Optional additional input specifications
```
osterman:
```suggestion
## - Optional additional input specifications
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+##   Ensure new definitions override defaults
```
osterman:
```suggestion
##   - Ensure new definitions override defaults
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+##   Optional suppression of default input specifications
```
osterman:
```suggestion
##   Optional suppression of default input specifications
```
```suggestion
##   - Optional suppression of default input specifications
```

rootfs/usr/local/bin/grafana-db
```diff
@@ -0,0 +1,161 @@
...
+fi
+
+if (($# != 1)); then
+	cat <<EOF
+Usage:
+  grafana-db <URL or gnetId>
+
+  URL is URL to (raw) ConfigMap or ConfigMapList containing Grafana dashboards
+  as for Prometheus Operator gnetId is ID (typically a 4 digit number) for the
+  dashaboard on Grafana.com
```
osterman: Please provide example usage.

Dockerfile
Nuru: These comments are what we should run when we switch to Python3. I left them there as documentation as it took me a while to figure out. 

I commented them out because it is not reasonable for us to switch to only Python 3 at this time, as neither `gcloud sdk` nor `crudini` support Python 3 at this time.

Dockerfile
Nuru: see above

Dockerfile
osterman: There's no imminent date as to when that will be. Copy & paste it to the PR as a comment for reference. Python3 is like Perl6. One day it will happen, just we don't know when. =)

(also, another reason I wish we didn't need the python dependencies altogether)

PR #515, opened 2019-08-23
Update label.yml

osterman:
## what
* Automatically label PRs based on what they modify

## why
* Easily filter PRs based on what they manage


osterman:
Apparently, this action is broken. See: https://github.com/actions/labeler/issues/6

PR #525, opened 2019-09-25
Feature: Toggle AWS Vault Helptext

osulli:
## What

Adds ability to pass an ENV in individual Geodesic module's Dockerfiles to toggle the prompt.

By default this will not affect anyone.

## Why

I am not a fan of the help text that is prompted on every command run. The big red cross / green tick is sufficient for me.

## Usage

Dockerfile
```
ENV AWS_VAULT_DISPLAY_PROMPT=false
```


goruha:
@osulli thanks for your pull request.
We will review and test it asap.

osulli:
Don't merge.
That is a reserved variable name.

osulli:
Fixed. No longer using variables from https://github.com/99designs/aws-vault/blob/master/USAGE.md#environment-variables

Nuru:
@osulli I don’t understand the point of this and don’t like adding another flag.

What is wrong with just setting `AWS_VAULT_ENABLED=false`?

osulli:
@Nuru 

Use case: Have Vault enabled in Geodesic (usable) but not have the prompt visible for those that do not use it (i.e. Azure AD users that auth through other means).

Alternatives suggested:
- Have a `make build` argument that reads from the engineer's `.bashrc`, this is fine also.. to save setup from others it could be placed in `.envrc` 

... but it still doesn't support 'AWS_VAULT_ENABLED' but also having the prompt disabled. The 'red cross' or 'green tick' are sufficient for seeing if authed and all my users know how to use aws-vault.

This is a backwards-compatible solution allowing further customization of Geodesic and I don't see how you can have it without another variable.

osulli:
I am closing this as it isn't going anywhere.

For reference if anyone else wants this:


```
└── rootfs
    ├── etc
    │   └── profile.d
    │       └── z_aws-vault.sh
```

```
#!/bin/bash

PROMPT_HOOKS+=("aws_vault_prompt")
function aws_vault_prompt() {
  if [[ -z "${AWS_VAULT}" ]] && [[ "${AWS_VAULT_DISPLAY_PROMPT}" != "false" ]]; then
    echo -e "-> Run '$(green assume-role)' to login to AWS with aws-vault"
  fi
}
```

Works great.

PR #534, opened 2019-11-01
add Darwin support for bind mount of ssh-agent socket

tamsky:
## what
Docker for Mac v 2.2.0.0 allows forwarding `ssh-agent` and `docker` Unix sockets to containers, allowing people to access the `ssh-agent` and `docker` agents on their Mac host from inside the Docker container. Previously, Geodesic automatically forwarded the `SSH_AUTH_SOCK` and optionally forwarded the Docker agent socket (enabled by `export WITH_DOCKER=true`) only on Linux hosts. With this release, that behavior is extended to Mac OS hosts when `export GEODESIC_MAC_FORWARD_SOCKET=true` is run before starting Geodesic.

## why
Among other things, this allows `ssh` credentials to be securely stored on the Mac host, taking advantage of the OS X Keychain, and still be used from inside the Docker container, for example to access a private Github or Docker repository. 

## original
@tamsky  wrote:

Newly available on Docker for Mac (Edge) release.
https://docs.docker.com/docker-for-mac/edge-release-notes/

Someone should test that this doesn't break on non-Edge Docker-for-Mac releases.


rootfs/templates/wrapper
```diff
@@ -68,19 +68,23 @@ function use() {
 					--env USER_ID
 					--env GROUP_ID)
 			fi
+			if [ "${OS}" == 'Darwin' ]; then
+				# Bind-mount SSH-agent socket (available in docker-for mac Edge release)
+				DOCKER_ARGS+=(--volume /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock
+					-e SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock)
```
Nuru: I do not like that this sets SSH_AUTH_SOCK even when `/run/host-services/ssh-auth.sock` does not exist, or that it mounts `/run/host-services/ssh-auth.sock` rather than the current value of `SSH_AUTH_SOCK`. What's wrong with following the pattern we use for Linux hosts?

osterman: The current pattern doesn’t work on Mac. He has solving it so it works on Docker for Mac .

Nuru:
> Newly available on Docker for Mac (Edge) release.
> https://docs.docker.com/docker-for-mac/edge-release-notes/
> 
> Someone should test that this doesn't break on non-Edge Docker-for-Mac releases.

It does, in fact break on other Docker releases.

tamsky:
Thank you @aknysh for the review, and @nuru for completing this!

PR #561, opened 2020-02-28
fix direnv

osterman:
## what
* Temporarily use `XDG_CONFIG_DIR`

## why
* https://github.com/direnv/direnv/issues/595

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -204,6 +204,10 @@ ENV MAKE_INCLUDES="Makefile Makefile.*"
 # Read more: <https://wiki.archlinux.org/index.php/XDG_Base_Directory>
 ENV XDG_CONFIG_HOME=/etc
 
+# This is a temporary fix related it https://github.com/direnv/direnv/issues/595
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -204,6 +204,10 @@ ENV MAKE_INCLUDES="Makefile Makefile.*"
 # Read more: <https://wiki.archlinux.org/index.php/XDG_Base_Directory>
 ENV XDG_CONFIG_HOME=/etc
 
+# This is a temporary fix related it https://github.com/direnv/direnv/issues/595
+# This can be removed on the next release of direnv > 2.21.2
```
Nuru:
```suggestion
# This is a temporary fix related it https://github.com/direnv/direnv/issues/595
# This can be removed on the next release of direnv > 2.21.2
# Note that XDG_CONFIG_DIR is not a variable mentioned in the XDG standard, 
# and should not be confused with XDG_CONFIG_DIRS.
```

PR #566, opened 2020-03-23
UI support for EKS, SAML auth, and Terraform Workspaces

Nuru:
## what
1. Add terraform workspace status to the command prompt when `GEODESIC_TERRAFORM_WORKSPACE_PROMPT_ENABLED=true`
1. Add `set_cluster` and `eks-update-kubeconfig` commands to facilitate connecting to and changing EKS Kubernetes cluster command targets (similar to `kubectx` but generates context files as well)
1. Update [`kube-ps1`](https://github.com/jonmosco/kube-ps1) to version v0.7.0, evolve our helper script, and add additional configuration for integration of Terrafrom workspace into prompt
1. When not using `aws-okta` or `aws-vault`, monitor AWS credentials to determine AWS role that would be used by default for an `aws` CLI command and export that to the command line prompt
1. Make `direnv` optional. Setting `DIRENV_ENABLED=false` in your Dockerfile will prevent the `direnv` hook from being installed. This can save considerable time if you are not using `direnv`.
1. Update `packages` MD5 to get latest cloudposse pacakges
1. Add comment about how to install [`aws-google-auth`](https://github.com/cevoaustralia/aws-google-auth). We do not install it by default because it is a python package with a lot of dependencies that adds more bloat than we are comfortable with for a tool few people will use.

## why

1. When using Terraform Workspaces, the currently active workspace is (a) critically important, as it determines which resource environment will be affected by actions, and (b) not immediately obvious. This change makes it obvious before every command execution. At the same time, it remains backwards-compatible with people not using workspaces.
1.  With `kops`-created clusters, `kops` also manged the `kubecfg` file and we have support for that workflow (e.g. `kexp`). With EKS-created clusters, we need a different workflow to manage `kubecfg` files, and `set_cluster` and `eks-update-kubeconfig` provide that. `set_cluster` is provided as a shell function so that it can manipulate the `KUBECONFIG` environment variable, while `eks-update-kubeconfig` does most of the work.
1. Take advantage of new feature that lets us shorted cluster names (the EKS cluster names are quite long). Use less resource-intensive scripting to handle hiding and showing the cluster in the prompt as `kube-ps1` has added capability we no longer need to provide.
1.  When using `aws-okta` or `aws-vault`, we explicitly track the current role based on hooks added to all the commands for manipulating roles. When using SAML SSO or other mechanisms for managing credentials, we do not have those hooks, so this PR adds monitoring of AWS credentials and updating the effective role when changes are detected. This should work with any credential provider to show the role the `aws` CLI will operate as by default.
1. `direnv` takes a noticeable amount of time to execute and it executes after every command. We are moving away from using it, so we can save time and unanticipated consequences by allowing it to be disabled.
1. There is always a danger of the package layer being cached by Docker if we do not update the packages MD5
1. Some people prefer  [`aws-google-auth`](https://github.com/cevoaustralia/aws-google-auth), but we do not install it by default because of the bloat, yet it is not trivial to install, so we provide instructions for others who might want to install it.


rootfs/etc/profile.d/set_cluster.sh
```diff
@@ -0,0 +1,20 @@
...
+#   and updates KUBECONFIG to point ot that file.
+
+#   With "off", deletes the currently active kubecfg file and unsets KUBECONFIG
+#
+
+function set_cluster() {
+	KUBECONFIG_DIR=$(dirname ${KUBECONFIG:-/dev/shm/kubecfg})
+	export EKS_KUBECONFIG_PATTERN="${EKS_KUBECONFIG_PATTERN:-${KUBECONFIG_DIR}/kubecfg.%s}"
+	if [[ $1 == "off" ]]; then
+		/Users/jgro/CP_dev/geodesic/rootfs/usr/local/bin/eks-update-kubeconfig "$@" && unset KUBECONFIG
```
osterman: This does not look deliberate =)

rootfs/usr/local/bin/eks-update-kubeconfig
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Usage:
+#   eks-update-kubeconfig <cluster-short-name>|off
```
osterman: Thoughts on calling it `export-eks-kubeconfig`?

osterman: Nevermind, I see what you gave it this name because the underlying command is `aws eks update-kubeconfig`

rootfs/etc/profile.d/set_cluster.sh
```diff
@@ -0,0 +1,20 @@
...
+# Usage:
+#   set_cluster <cluster-short-name>|off
+#
+#   With <cluster-short-name> updates the kubecfg file for the cluster with that short name (e.g. "corp")
+#   and updates KUBECONFIG to point ot that file.
+
+#   With "off", deletes the currently active kubecfg file and unsets KUBECONFIG
+#
+
+function set_cluster() {
```
osterman:
```suggestion
function set-cluster() {
```

rootfs/etc/profile.d/set_cluster.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Usage:
+#   set_cluster <cluster-short-name>|off
```
osterman:
```suggestion
#   set-cluster <cluster-short-name>|off
```

rootfs/etc/profile.d/set_cluster.sh
osterman: there's no prior art in geodesic of calling commands with `_`

Nuru: Standard is all shell functions use underscores because `-` is not a valid `bash` identifier character. This has to be a shell function in order to manipulate the caller's environment.

Nuru: We use underscores in all the shell function names. I didn't even realize dashes were allowed (they are not allowed in variable names).

PR #608, opened 2020-08-09
Better support for XDG, update Helm plugins

Nuru:
## what
- Set XDG variables to share "per-user" configuration
- Update Helm plugins

## why
- Geodesic is meant to be used by only one "user" at a time. Docker builds its image as the root user, so by default, installations (such as Helm) that respect the XDG standard install items under `~root`. However, if we want to run a daemon in Geodesic as a non-privileged user (e.g. Atlantis), we still want that user to have access to whatever tools we installed. To promote this, as well as to keep track of which packages are referencing XDG, we explicitly set the XDG variables to shared folders in standard directories. 
- Bug fixes

## references
- [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
- [ArchUnix Wiki entry on XDG](https://wiki.archlinux.org/index.php/XDG_Base_Directory)
- [How Helm uses XDG](https://helm.sh/docs/faq/#xdg-base-directory-support)


Dockerfile
```diff
@@ -172,6 +185,11 @@ RUN helm3 plugin install https://github.com/databus23/helm-diff.git --version v$
     && helm3 plugin install https://github.com/aslafy-z/helm-git.git --version ${HELM_GIT_VERSION} \
     && helm3 plugin install https://github.com/helm/helm-2to3 --version ${HELM_HELM_2TO3_VERSION}
 
+# Delete unneeded Helm plugin cache
```
osterman, line 188: Fwiw, since this is a different `RUN` step, the previous layer will still contain the cache so it will not affect the final size of the image.

Nuru: I totally forgot about that. Let me fix.

PR #616, opened 2020-08-30
Add all kubectl versions, add auto-selector. Update packages.

Nuru:
## what

- Switch from including `kubectl` package, which installs latest kubectl (currently v1.19.0) to including only `kubectl-1.16` package, which is the most compatible with the versions actually in use by our customers
- Add `kubectl-auto-select` script that selects and optionally installs the correct `kubectl` given a connection to the server, or based on command line input
- Update to latest `cloudposse/packages`
- Update `.dockerignore` to ignore more files and directories

## why

- The version of `kubectl` must closely match the version of the Kubernetes control plane it is operating with. We were installing version 1.19 which is cutting edge and not compatible with Kubernetes 1.17, while most people are using Kubernetes 1.15, 1.16, or 1.17
- Give people some help installing the right version for their system.
- Critical tools (helm, helmfile, terraform) are being updated quickly 
- Smaller Docker images are good for everybody


rootfs/usr/local/bin/kubectl-auto-select
```diff
@@ -0,0 +1,84 @@
...
+}
+
+# Set Kubectl version based on version running on server, if available,
+# otherwised based on environment variable, if set
+function set_from_server() {
+	local versions clientVersion serverVersion hasServerVersion
+
+	versions=$(kubectl version -o json 2>/dev/null)
+	hasServerVersion=$?
+	clientVersion=$(normalize_version $(jq -r .clientVersion.minor <<<$versions))
```
3h4x, line 75: This will stop working with kubectl 2.

rootfs/usr/local/bin/kubectl-auto-select
```diff
@@ -0,0 +1,84 @@
...
+	if (($hasServerVersion != 0)); then
+		set_from_envar $clientVersion
+	else
+		serverVersion=$(normalize_version $(jq -r '.serverVersion.minor' <<<$versions))
+		set_alternatives "$serverVersion"
+	fi
+}
+
+set_from_server
```
3h4x, line 84: A bit misleading name of a function as `set_from_server` might set a version from env variable if `server` version not available.
I'd create function `detect_version` with that if logic:
```
	if (($hasServerVersion != 0)); then
		set_from_envar $clientVersion
	else
		serverVersion=$(normalize_version $(jq -r '.serverVersion.minor' <<<$versions))
		set_alternatives "$serverVersion"
	fi
```

rootfs/usr/local/bin/kubectl-auto-select
```diff
@@ -0,0 +1,84 @@
...
+	if [[ $1 == "null" ]]; then
+		red "Unable to determine Kubernetes (kubectl) version"
+		exit 1
+	fi
+
+	local version=${1#v} # remove leading "v" if present
+	# Customized Kubernetes versions, such as from AWS EKS, may have more than just the
+	# digits in the minor version, e.g. "16+", so strip out everything after leading digits,
+	# including patch version.
+	version=$(sed -E 's/((\d\.)?(\d+))([^d].*)?/\1/' <<<$version)
```
3h4x, line 25: I think this function should accept input like `v1.16.whatever`.That way at any time we are dealing with whole "version" string and not just part of it (like minor). It would simplify later on getting version from kubectl:
`kubectl version -o json | jq .serverVersion.gitVersion -j`

If so then I'd use this regex to parse it:
`v?(\d{1}\.\d{2}).*` which seems to be much more clear. 

Or `\d+` to match also future kubectl versions like 1.100 (if it will ever exist).

3h4x (approved):
Good stuff!
I left nonblocking comments.

rootfs/usr/local/bin/kubectl-auto-select
```diff
@@ -0,0 +1,84 @@
...
+
+# Use `update-alternatives` to set kubectl to given major.minor version
+function set_alternatives() {
+	local version="$1"
+
+	if update-alternatives --set kubectl /usr/share/kubectl/"${version}"/bin/kubectl-"${version}"; then
+		if [[ $version == $(kubectl_mm_version) ]]; then
+			green "kubectl version set to $(kubectl_git_version)"
+		else
+			red "Unable to set kubectl to version ${version}"
```
osterman, line 48:
```suggestion
			red "Unable to set kubectl to version ${version}. Perhaps try installing it with \`apk add kubectl-${version}@cloudposse\`?"
```

rootfs/usr/local/bin/kubectl-auto-select
Nuru: Yes, we have a lot of stuff that will stop working with Kubernetes 2.x. I prefer to address it when that release actually happens, as who know what else will have changed between now and then.

rootfs/usr/local/bin/kubectl-auto-select
Nuru: Your version is not that different than mine. The key difference is mine accepts `16` as well as `v1.16.whatever`.

PR #628, opened 2020-10-20
[grafana-db] Add --overwrite flag, [aws-region] add an3 region

Nuru:
## what

- [grafana-db] Add --overwrite flag
- [aws-region] add ap-northeast-3 => an3 region 
- Replace `man-db` package with `man`

## why

- Allow scripts to update/replace standard dashboards even if they have been modified
- Asia Pacific (Osaka-Local) `ap-northeast-3` region not previously included
- Alpine packages for `man-db` have dropped support for Alpine 3.11, which is what we are using. Replacing with still-supported `man`


rootfs/usr/local/bin/grafana-db
```diff
@@ -185,6 +188,17 @@ function _main() {
...
 		return 4
 	fi
+	case $1 in
+	--overwrite)
+		OVERWRITE=true
+		shift
+		;;
+
+	--*)
+		red Unknown option "$1"
```
aknysh, line 198:
```suggestion
		read Unknown option "$1"
```

aknysh (approved):

aknysh (approved):

Issue #615, opened 2020-08-27
General Enhancements


osterman:
## what
- Only use `--privileged` mode when using `aws-vault` (see https://github.com/cloudposse/geodesic/pull/143) or as needed. but we should default to not using it so files are not owned by `root`. #139 
- Support a feature flag so we do not reattach to existing session

Nuru:
I am closing this:

- Getting rid of the `--privileged` flag appears to be asking too much. We might be able to reduce it to `--cap-add SYS_ADMIN`, but that is not much of an improvement. So closing as `wontfix` but you can open a separate, isolated issue to discuss this if you want to.
- #961 adds a feature flag so we do not reattach to existing session

PR #638, opened 2020-11-13
Create Debian-based version

Nuru:
## what

- Create a new Dockerfile to use to create a Geodesic image based on Debian buster (10.6) rather than Alpine 3.11
- Update Makefile to make and install user's choice of Debian or Alpine-based Geodesic
- Use GitHub Actions to build Docker images. Closes #623 
- Install Feature Request and Bug Report templates. Closes #482 
- Add CODEOWNERS file

## why

- Debian is based on the full Linux kernel, unlike Alpine, which makes it more broadly compatible. In particular, [Alpine does not have GNU `glibc`](https://wiki.alpinelinux.org/wiki/Running_glibc_programs), which makes it incompatible with a number of programs. 

## notes

The startup routine now displays the base OS of the image (above the banner).

Make targets build the Alpine version by default, but now the Makefile supports a `DOCKER_BASE_OS` environment variable and targets like `debian.build` to switch between Alpine and Debian versions. By default the Docker images get tagged with both `dev` and `dev-$DOCKER_BASE_OS` and `make install` installs a script which references and runs the `dev-$DOCKER_BASE_OS` tag. 

Still some kinks to work out, like the fact that the container is still named `geodesic` and the script will attach to the running container regardless of the tag (see #615) but fortunately few people actually build and run Geodesic directly. More of an issue when adapting the Geodesic Makefile to your custom Makefile.


osterman (dismissed):

README.md
```diff
@@ -91,77 +102,74 @@ Let's roll...
...
 4. Quickly provision [kops clusters](docs/kops.md)
 
+## Usage
 
 
 
+In general we recommend creating a customized version of Geodesic by creating your own Dockerfile starting with
+```
+ARG VERSION=0.138.0
+ARG OS=debian
```
osterman, line 112: `BASE_OS` to match docs?

osterman (approved):

PR #653, opened 2020-12-13
Use release-drafter to draft release notes

Nuru:
## what
- Use release-drafter to automatically generate release notes combining multiple automatic upgrades and manual changes

## why
- More consistent release notes more easily including all automated updates in a single release

## notes
- This does not automatically create a release, just draft the release notes. Releases are still done manually, and release notes and version numbers can be edited before release.

### DIFF
--- a/.github/draft-release.yml
+++ b/.github/draft-release.yml
@@ -1,0 +1,61 @@
+name-template: 'v$RESOLVED_VERSION'
+tag-template: '$RESOLVED_VERSION'
+version-template: '$MAJOR.$MINOR.$PATCH'
+version-resolver:
+  major:
+    labels:
+    - 'major'
+  minor:
+    labels:
+    - 'minor'
+    - 'enhancement'
+  patch:
+    labels:
+    - 'auto-update'
+    - 'patch'
+    - 'fix'
+    - 'bugfix'
+    - 'bug'
+    - 'hotfix'
+  default: 'minor'
+
+categories:
+- title: '🚀 Enhancements'
+  labels:
+  - 'enhancement'
+  - 'patch'
+  - 'shell'
+  - 'scripts'
+  - 'terraform'
+- title: '🐛 Bug Fixes'
+  labels:
+  - 'fix'
+  - 'bugfix'
+  - 'bug'
+  - 'hotfix'
+- title: '🤖 Included Tools'
+  labels:
+  - 'auto-update'
+  - 'packages'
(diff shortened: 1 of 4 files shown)


osterman (changes requested):
Please use GITHUB_TOKEN instead like our other release actions and rename the config for consistency.

Nuru:
> Please use GITHUB_TOKEN instead like our other release actions and rename the config for consistency.

@osterman All our `auto-release` workflows use `PUBLIC_REPO_TOKEN`. See the template at [build-harness/templates](https://github.com/cloudposse/build-harness/blob/master/templates/.github/workflows/auto-release.yml). This action is exactly the template action except here I set `publish: false`.

I will change the names if you want. My thinking was that since we use `auto-release` to automatically make releases, and this is not doing that, I should name it differently in order to avoid confusion. My fear is that if it is named "auto-release" people will see that and think that they do not need to manually cut a release.

osterman:
@Nuru that changed recently and must have slipped through code review. I would prefer we use the built-in tokens. 

image

osterman:
The version in build-harness that I added had [`GITHUB_TOKEN`](https://github.com/cloudposse/build-harness/commit/902ff51c1c85e647c925c31daa658ee7a5317466#diff-2e748c8595cb97c13b6bfe146fc2f026ac827748805ca3302d3c47ce8805e5f6L17), then it was changed in https://github.com/cloudposse/build-harness/pull/249 without explanation.

Nuru:
> The version in build-harness that I added had [`GITHUB_TOKEN`](https://github.com/cloudposse/build-harness/commit/902ff51c1c85e647c925c31daa658ee7a5317466#diff-2e748c8595cb97c13b6bfe146fc2f026ac827748805ca3302d3c47ce8805e5f6L17), then it was changed in [cloudposse/build-harness#249](https://github.com/cloudposse/build-harness/pull/249) without explanation.

@osterman It was copied from [here](https://github.com/cloudposse/pritunl/pull/7), with the explanation:

> osterman:
> Aha so one GitHub action can not trigger another GitHub action for normal events unless using the PUBLIC_REPO_ACCESS_TOKEN
>
> So that explains why the auto release was not triggering build on release created

And later, regarding the use if "v" in version numbers, but regarding the same PR and auto-release action
> osterman:
> unfortunately, this should be fixed in the upstream `build-harness` template which we copy everywhere

osterman (approved):

PR #654, opened 2020-12-14
[draft-release] Fix broken syntax in config file

Nuru:
## what
- Fix broken syntax in `draft-release` config file

## why
- Action would not execute because of syntax errors


mcalhoun (approved):
LGTM

osterman (approved):

PR #664, opened 2020-12-20
Limit AWS CLI updates to once a week

Nuru:
## what
- Limit `aws` CLI and `boto3` updates to once a week and group them together

## why
- Reduce the noise. In general these are updated several times a week but the updates are not urgent.
- Group them because they are typically released together (since `aws` depends on `boto3`) and `aws` pins `boto3` in such a way that the `boto3` update would break `aws` if `aws` is not updated at the same time.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,6 +1,14 @@
 {
+  "timezone": "America/Los_Angeles",
   "extends": [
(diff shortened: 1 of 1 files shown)


osterman (approved):
LGTM

PR #693, opened 2021-03-14
Tweaks for various "docker run" misconfigurations

Nuru:
## what
- Fix some typos from #692 
- Add instructions at bottom of installation script for when all else fails
- change `DOCKER_TAG` default from dev to whatever the tag is based on the docker build step in the workflow
- fix `docker run -it .... init | bash` from failing due to `\r` 
- Standardize bullets to use `#` during startup 

## why
- bugfix
- Improve experience when Docker images derived from Geodesic are run incorrectly
- Try to reference a Docker image that exists
- Some people type `docker run -it` out of habit
- Aesthetics/UX


osterman (dismissed):

osterman:
## todo
- [ ] change `DOCKER_TAG` default from dev to whatever the tag is based on the docker build step in the workflow
- [ ] fix `docker run -it .... init | bash` from failing due to `\r` image

rootfs/usr/local/bin/init
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# If output is a TTY, turn off conversion of NewLine to CarriageReturn-NewLine
+# (If output is not a TTY, `stty` will complain about that)
+[[ -t 1 ]] && stty -onlcr
```
osterman, line 5: nice

osterman (approved):

PR #720, opened 2021-06-30
feat: updates mixed prompt to be warning > error

Gowiem:
## What

* Updates aws role prompt to not show `!mixed!` when the `aws-vault` profile name does not match the assumed user or role.
* Updates to use yellow > red as this is a warning and not an actual error

## Why

* The aws-vault profile name not matching up to the assumed user or role is not an error event: For myself (and others I'm sure), it is an expected outcome of utilizing the same AWS IAM user username (e.g. `matt.gowie` in my instance) across many accounts / clients and needing to name those various accounts separately in `aws-vault` so I can differentiate which account I'm utilizing.

## Additional Info

* Previously this looked like: 
CleanShot 2021-06-30 at 10 09 13

* With this change, it now looks like what I would expect and prefer: 
CleanShot 2021-06-30 at 10 08 18

### DIFF
--- a/rootfs/etc/profile.d/aws.sh
+++ b/rootfs/etc/profile.d/aws.sh
@@ -86,13 +86,14 @@ function export_current_aws_role() {
 	local profile_target=${AWS_PROFILE:-${AWS_VAULT}}
 	if [[ -n $profile_target ]]; then
 		profile_arn=$(aws --profile "${profile_target}" sts get-caller-identity --output text --query 'Arn' 2>/dev/null | cut -d/ -f1-2)
-		if [[ $profile_arn == $current_role ]]; then
-			export ASSUME_ROLE="$profile_target"
-			return
+
+		# If the profile name and the role / user we're assuming do not match then warn the user.
+		if [[ $profile_arn != $current_role ]]; then
+			echo "* $(yellow Profile is set to $profile_target but current role does not match:)"
+			echo "*   $(yellow $current_role)"
 		fi
-		echo "* $(red Profile is set to $profile_target but current role does not match:)"
-		echo "*   $(red $current_role)"
-		export ASSUME_ROLE=$(red '!mixed!')
+
+		export ASSUME_ROLE="$profile_target"
 		return
 	fi
 


Nuru (changes requested):
@Gowiem Thank you for pointing out this issue. However, I think you misunderstand the problem and why it is marked as a Red Error.

I believe it has nothing to do with you reusing your username across multiple accounts. 

The fundamental issue is that Geodesic expects the AWS_VAULT profile to match the AWS_PROFILE profile and expects the AWS_PROFILE (and therefore the AWS_VAULT) to be in your AWS_CONFIG_FILE. Is there a reason this is impossible or undesirable? 

Specifically, the test that is failing is that (shortened from actual command)
``` 
aws sts get-caller-identity
```
is returning a different identity than
``` 
aws --profile bonkers sts get-caller-identity
```
which will cause uncertainty about which role to use for execution. That is why it is a Red Error. Which role will be used in various Geodesic scripts will depend on whether or not they explicitly reference AWS_PROFILE or AWS_VAULT.

Gowiem:
@Nuru got it -- Thanks for clarifying things. This still seems problematic to me, so let me show you a full reproduction of the issue: 

CleanShot 2021-07-08 at 15 06 10@2x

I login utilizing `aws-vault exec <profile> -- bash --login` because utilizing the `assume-role` helper consistently hides my MFA and / or aws-vault password prompts. When I do that, it immediately invokes the prompt function but because I'm already in an `aws-vault` session (and therefore `AWS_VAULT` is set) the `aws --profile bonkers sts get-caller-identity` command fails with the following: 


> Error when retrieving credentials from custom-process: aws-vault: error: exec: aws-vault sessions should be nested with care, unset $AWS_VAULT to force

This is because my AWS Config utilizes `credential_process`, and looks like the following: 

```
[profile bonkers]
mfa_serial=arn:aws:iam::<Account ID>:mfa/matt.gowie
credential_process=aws-vault exec bonkers --json
region=us-west-2
output=json
```

How is this not happening for you / others? Are you not utilizing `credential_process`? What am I missing here?

Nuru:
@Gowiem No, we do not use `credential_process` and we do not recommend using it. AFAIK the makers of `aws-vault` do not recommend using `aws-vault` as a credential provider, either. 

At this point, we do not recommend using `aws-vault` either, partly because of the issues you cite and things like [not hiding the password](https://github.com/99designs/aws-vault/issues/689). If you want to use `aws-vault` with Geodesic, we recommend using version 4.7.2, running a credential server in the foreground with the Geodesic helper

```
role-server bonkers
```

and then running a second shell in the same Geodesic container to do your work. When you run Geodesic via `make run` or by using the provided wrapper script, if the container is already running you will get a second shell in the same container rather than a new instance of the container. With the credential server running, the second shell will get the credentials it needs for the profile you set when you ran the server. When the server needs a new MFA token, the other shells will error out due to lack of valid credentials, but since you ran the server in the foreground, its prompt for MFA token will be visible and you can enter the new token without problems.

In any case, you should remove the `credential_helper` line from your profile. It just doesn't work and is not necessary.

You may also be interested in using [Leapp](https://www.leapp.cloud/) instead of `aws-vault`. It is new and has warts and issues, but it is being actively developed and has a decent set of capabilities at this point. We are not yet ready to fully recommend it due to some of the outstanding issues (such as not being able to export and import configurations), but I expect that the developers will address the issues and we will be recommending Leapp in the not too distant future. 

Once you stop using `aws-vault` and set `AWS_VAULT_ENABLED=false` in your Dockerfile, you can start taking advantage of the recent improvements in the AWS SDK for role chaining and pretty much give up needing to use `assume-role`. You just set your `AWS_PROFILE` to what you want it to be. It is actually a _lot_ simpler now that AWS has distributed and tool makers are using an SDK that fully supports all the various credential sources and supports automatic role chaining. We stopped using `aws-vault` on new client engagements quite a while ago.

Gowiem:
@Nuru I'm not seeing `role-server` in geodesic either locally, in `master`, or in `cloudposse/packages`. Is that hidden somewhere I don't know about?

Regardless, I'm trying to follow your process and still running into issues. When trying to utilize the `aws-vault exec bonkers --server` and I'm getting the following: 

> aws-vault: error: exec: fork/exec : no such file or directory

[Related to some python / aws issue with aws-vault](https://github.com/99designs/aws-vault/issues/624).

Regarding your mention of not utilizing aws-vault at all: You're 100% on SAML and / or AWS SSO, correct? I think that is one of the major differences between our setups. Since I've been utilizing `aws-vault` I have been against putting any creds in `~/.aws/credentials`, but I feel like you're pointing that direction which I'm surprised about from a security perspective. What are your thoughts there?

One thing in that vein regarding your mention of

> you can start taking advantage of the recent improvements in the AWS SDK for role chaining and pretty much give up needing to use assume-role.

What prompts you for MFA in this scenario? Does the newer CLI now support that?

Nuru:
@Gowiem 

> @Nuru I'm not seeing `role-server` in geodesic either locally, in `master`, or in `cloudposse/packages`. Is that hidden somewhere I don't know about?

`role-server` is a [shell function](https://github.com/cloudposse/geodesic/blob/0616dec70e2a11dd66aca2642407c5093a2b79b8/rootfs/etc/profile.d/aws-vault.sh#L280) set up in `aws-vault.sh`
> 
> Regardless, I'm trying to follow your process and still running into issues. When trying to utilize the `aws-vault exec bonkers --server` and I'm getting the following:
> 
> > aws-vault: error: exec: fork/exec : no such file or directory
> 
> [Related to some python / aws issue with aws-vault](https://github.com/99designs/aws-vault/issues/624).
> 

That seems to be an issue with aws-vault v6.x. Again I recommend you use v4.7.2 inside Geodesic if you are going to use it at all. (Unless you are using `kops`, which has issues with `aws-vault` server mode and in which case I again recommend dropping  `aws-vault` entirely.)

> Regarding your mention of not utilizing aws-vault at all: You're 100% on SAML and / or AWS SSO, correct? 

No, I still use plain API keys in some places.

> I think that is one of the major differences between our setups. Since I've been utilizing `aws-vault` I have been against putting any creds in `~/.aws/credentials`, but I feel like you're pointing that direction which I'm surprised about from a security perspective. What are your thoughts there?

While I remain firmly opposed to keeping any long-lived credentials stored unencrypted on disk, with AWS it is just impractical to support multiple accounts and multiple programs with environment variables alone. The AWS SDK itself caches short-lived credentials on disk (in `~/.aws/cli/cache/` and `~/.aws/sso/cache/`) and so I support saving ***short-lived*** credentials in `~/.aws/credentials`. My view is that if an adversary is able to exfiltrate your credentials file in under a few hours, they can most likely also (and probably even more easily) exfiltrate the environment variables of any process on your machine.  

So all the tools I use now (`aws`, `aws-google-auth`, `saml2aws`, Leapp) all store short-lived session credentials in a file and that file is available (generally via the AWS SDK) to any tools than need AWS access. Long-term credentials are stored encrypted somewhere or memorized. 

> 
> One thing in that vein regarding your mention of
> 
> > you can start taking advantage of the recent improvements in the AWS SDK for role chaining and pretty much give up needing to use assume-role.
> 
> What prompts you for MFA in this scenario? Does the newer CLI now support that?

As with `aws-vault`, MFA is only used when converting long-lived credentials into short-lived credentials, and that prompt happens when the conversion is made (session is created) by the tool making the conversion.

Gowiem:
@Nuru got it and again thanks for the knowledge and break down. I'll close this out.

Issue #726, opened 2021-08-16
Start container in pwd


nitrocode:
Have a question? Please checkout our [Slack Community](https://slack.cloudposse.com) or visit our [Slack Archive](https://archive.sweetops.com/). 

[![Slack Community](https://slack.cloudposse.com/badge.svg)](https://slack.cloudposse.com)

## Describe the Feature

Start geodesic in the pwd where the container was run from

## Use Case

I always cd into /localhost until i get to my pwd which isnt a big deal to do but it takes time and id prefer if the container automatically did this for me without updating preferences.

## Describe Ideal Solution

pwd

## Alternatives Considered

Maintaining a long list of aliases that cd into common directories

## Additional Context
N/A

Nuru:
@nitrocode Part of the point of Geodesic is to provide a consistent environment so multiple engineers can use the exact same tools and workflows and avoid the "works in my environment" trap. That is why the default working directory is inside the container. 

That said, we have made Geodesic customizable, so you can do what you want this way (among other options):
```
$ GEODESIC_DOCKER_EXTRA_ARGS="--env LOCAL_PWD=$PWD" make run
...

 ⧉  Geodesic 
 ✗ . [none] / ⨠ cd $LOCAL_PWD 
```

And of course you can automatically execute that command, too. For example, add to `$HOME/.geodesic/defaults/overrides` 
```
[[ -n $LOCAL_PWD ]] && cd $LOCAL_PWD 
```

nitrocode:
Re-opening this ticket since @osterman said last Friday during the architecture meeting that he wants to see this implemented.

Nuru:
Implemented in #756

PR #753, opened 2022-01-10
DEV-105: Set workdir to pwd

nitrocode:
## what
* Add workdir setting and default it to the present working directory from where the container was started
* Implementation works without gnu realpath
* Implementation works with or without gnu sed
* Implementation works even if `GEODESIC_DOCKER_EXTRA_ARGS` is set.

## why
* The benefit with this is that we start geodesic, we're automatically dropped in a path with in the container relative to our current working directory on the native filesystem
* This is a common configuration across customers and other toolbox implementations like the fedora toolbox.
* Original solution was to update `GEODESIC_DOCKER_EXTRA_ARGS` in the `Makefile` which only runs when running `make run|build|all` and not when running the wrapper directly.

## references
* Closes https://github.com/cloudposse/geodesic/issues/726
* Closes https://github.com/cloudposse/geodesic/pull/754
* DEV-105

## commands

```shell
$ echo ${GEODESIC_WORKDIR:-/localhost/${PWD#$HOME/}}
/localhost/git/github/cloudposse/geodesic/rootfs/templates
```

```shell
$ unset ATMOS_BASE_PATH
$ source atmos.sh
Set ATMOS_BASE_PATH = .
$ source atmos.sh
```


leb4r (approved):
Can confirm on debian and alpine that when entering the container via the wrapper, you're in the $PWD relative to `/localhost` which is mounted to `$HOME` . 

Is there a reason why there is a trailing `/` specifically at the end of `GEODESIC_WORK_DIR` ?

rootfs/templates/wrapper
```diff
@@ -6,6 +6,9 @@ set -o pipefail
 # Geodesic Settings
 export GEODESIC_PORT=${GEODESIC_PORT:-$((30000 + $$ % 30000))}
 
+# If this env var is unset, set it to the current directory without the home dir
+export GEODESIC_WORKDIR=${GEODESIC_WORKDIR:-/localhost/$(echo "$PWD/${1#./}" | sed -e "s,^$HOME/,,g")}
```
osterman, line 10: lets add a to `rootfs/etc/init.d/atmos.sh` which contains atmos specific enhancements.

Then let's set `export ATMOS_BASE_PATH=${GEODESIC_WORKDIR}`

osterman: maybe we should do something like
```
if [ -d "${GEODESIC_WORKDIR}/stacks" ] && [ -d "${GEODESIC_WORKDIR/components"] ; then
  export ATMOS_BASE_PATH=${GEODESIC_WORKDIR}
fi
```

aknysh: this will still not allow running `atmos` commands from any folder in the container, but should be a good start

osterman: > this will still not allow running atmos commands from any folder in the container, but should be a good start

This should work when the container was _started_ in a atmos root in the local filesystem.

rootfs/etc/init.d/atmos.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+if [ -d "${GEODESIC_WORKDIR}/stacks" -a -d "${GEODESIC_WORKDIR}/components" ]; then
```
osterman, line 3: Maybe check that `ATMOS_BASE_PATH` is not set?

rootfs/templates/wrapper
osterman, line 10: Bash natively supports prefix substitution (So no `sed` needed)

https://github.com/cloudposse/geodesic/pull/754/files#diff-e96fd8eb67742b91673eb204117f0efe4f6c1e2a5488d4a754a06e08570e6c5eR161

osterman (approved):

rootfs/templates/wrapper
```diff
@@ -6,6 +6,9 @@ set -o pipefail
 # Geodesic Settings
 export GEODESIC_PORT=${GEODESIC_PORT:-$((30000 + $$ % 30000))}
 
+# If this env var is unset, set it to the current directory without the home dir
+export GEODESIC_WORKDIR=${GEODESIC_WORKDIR:-/localhost/${PWD#$HOME/}}
```
elventear, line 10: I think this default will not make sense if `$PWD` is outside of `$HOME`. That is why in #754 the workdir is set conditionally.

osterman: @nitrocode

nitrocode: The worst thing about this is if you started in a non home directory, you'd simply be placed in `/localhost` which seems acceptable.

PR #754, opened 2022-01-11
feat(wrapper): set workdir in docker similar to cwd at startup

elventear:
If $HOME is mounted into docker and the CWD is the inside $HOME, set the
docker workdir to be the same directory used when the wrapper was
called.


nitrocode:
Thanks @elventear for the PR. We have another PR to implement the same but to also allow additional env vars set for ATMOS.

https://github.com/cloudposse/geodesic/pull/753

elventear:
Closed in favor of #753

PR #756, opened 2022-01-21
Start in host CWD, autoconfigure ATMOS_BASE_PATH

Nuru:
## what
- When starting Geodesic, set the initial working directory to the Host's current working directory, if it is accessible
- If the initial working directory appears to be part of an `atmos` configuration, set `ATMOS_BASE_PATH` accordingly
- Allow fine-grained customization of command line prompt
- Update `kubectl` 1.16 -> 1.21
- Update Google Cloud SDK 352.0.0 -> 369.0.0
- Update Debian 10.10 -> 10.11
- Update Python3 on Debian 3.8.11 -> 3.8.12 
- Update Alpine 3.13.5 -> 3.13.7 (Python3 3.8.10 dictated by Alpine package repo)

## why
- Popularly requested features
- Outdated `kubectl` should have been upgraded long ago
- Security updates

## references
- closes and supersedes #755 
- replaces #753, #754
- https://github.com/cloudposse/atmos/pull/100
- Augments response to #417 

### Customizing command line prompt appearance

You could previously choose between 3 sets of glyphs to be used in the Geodesic command line prompt by [setting the environment variable](https://github.com/cloudposse/geodesic/blob/master/docs/customization.md) `PROMPT_STYLE` to "plain" to use plan ASCII characters, "unicode" to use Unicode characters limited to the standard (not supplemental) planes of Unicode, or leave it unset (or set to anything else) to get the default set of glyphs, which notably includes '⨠' Z NOTATION SCHEMA PIPING from the "Supplemental Mathematical Operators" Unicode block. Being from the supplemental plane, it is not as widely available: for instance, it is not included in the default fonts on Ubuntu.

Our current recommendation for resolving this issue is to install a font that contains the missing character on your host. For example, for Ubuntu:
```bash
$ sudo apt update
$ sudo apt install fonts-noto-core
```

However, taking this as a general issue rather than just specific to Ubuntu, we have included more options.

**New PROMPT_STYLE:** You can now set `PROMPT_STYLE=fancy` to get the default set of glyphs except for the one missing from Ubuntu, which is replaced with the glyph from the "unicode" style. This should cover most cases where the default style is broken.

**GEODESIC_PROMPT_GLYPHS:** We previously provided a specific workaround for this by allowing you to set `GEODISIC_PROMPT_GLYPHS` [sic] to any string of characters you wanted to end the prompt with. In this release we fixed the spelling of that environment variable, so it is now `GEODESIC_PROMPT_GLYPHS`, but hopefully you will no longer need it, and you will instead choose either to install a font or switch to the "fancy" style.

**Full glyph customization:** You can now set each of the 4 glyphs used by the command line prompt, plus the host file system designator, individually:
image

- `ASSUME_ROLE_ACTIVE_MARK` is the glyph to show when you have AWS credentials active. Defaults to a green, bold, '√' SQUARE ROOT (looks like a check mark): `$'\u221a'`
- `ASSUME_ROLE_INACTIVE_MARK` is the glyph to show when you do not have AWS credentials active. Defaults to a red, bold, '✗' BALLOT X: `$'\u2717'`
- `BLACK_RIGHTWARDS_ARROWHEAD` is the glyph at the end of the prompt. The troublesome default is '⨠' Z NOTATION SCHEMA PIPING: `$'\u2a20'`
- `BANNER_MARK` is the glyph at the start of the first line of a 2-line prompt that introduces the `BANNER` text. Defaults to '⧉', TWO JOINED SQUARES: `$'\u29c9'`
- `PROMPT_HOST_MARK` is added to the command line when the current working directory is on the host computer (via a bind mount) and not in the container. Defaults to '(HOST)' with "HOST" in red bold. Disable this feature by setting `export PROMPT_HOST_MARK=""`.

You can now set any of these in your [standard customizations](https://github.com/cloudposse/geodesic/blob/master/docs/customization.md) and your setting will be used. 

_Cautions_: 
- You can set these variables to multiple characters, and use ANSI escape codes to change their colors, but in order for command line editing to continue to work properly, any non-printing characters must be escaped. We have had the best luck with starting the escape with `$'\x01'` and ending it with `$'\x02'`, but there a couple of other, similar options. If you fail to do this, or do it incorrectly, your cursor will be incorrectly positioned when you edit a command line from your history and what you see will not be what is executed.
- Command line editing can also be affected by ["ambiguous width characters"](https://www.unicode.org/reports/tr11/tr11-39.html). Unicode characters can be "narrow" or "wide" or "ambiguous". In practice in Roman (e.g. English) scripts, a "wide" character takes up the space of 2 "narrow" characters, and the standard Roman letters are all narrow. In older versions of Unicode, many Emoji are ambiguous width. If you use an ambiguous width character that prints wide but is interpreted as narrow, command line editing will suffer from incorrect cursor placement, as with non-printing characters above. This can be worked around by finding and selecting a preference such as "Treat ambiguous characters as wide" (if you can find such a preference), but we recommend just avoiding characters that cause problems.

### DIFF
--- a/rootfs/etc/profile.d/_workdir.sh
+++ b/rootfs/etc/profile.d/_workdir.sh
@@ -21,7 +21,7 @@ else
 fi
 
 function file_on_host() {
-  [[ $GEODESIC_LOCALHOST_MISSING != "true" ]] && [[  $(_file_device "$1") == ${GEODESIC_LOCALHOST_DEVICE} ]]
+	[[ $GEODESIC_LOCALHOST_MISSING != "true" ]] && [[ $(_file_device "$1") == ${GEODESIC_LOCALHOST_DEVICE} ]]
 }
 
 function _default_initial_wd() {
--- a/rootfs/etc/profile.d/_workdir.sh
+++ b/rootfs/etc/profile.d/_workdir.sh
@@ -1,0 +1,60 @@
+# Files in the profile.d directory are executed by the lexicographical order of their file names.
+# This file sets the working directory inside Geodesic to match the host directory Geodesic
+# was launched from, if possible. If the host directory is not accessible, it sets the working directory to `/`.
+#
+# This file is named _workdir.sh. The leading underscore is needed to ensure this file executes before
+# other files that may depend on it. The "w" is needed to ensure it is loaded *after* _preferences.sh
+#
+
+function _file_device() {
+	df --output=source "$1" | tail -1
+}
+
+# file_on_host is true when the argument is a file or directory that appears to be on the Host file system.
+# Intended to support files on user-defined bind mounts in addition to `/localhost`.
+# This function is run by the command line prompt setup, so it should be very fast.
+# Therefore we cache some info in the environment.
+if df -a | grep -q /localhost; then
+	export GEODESIC_LOCALHOST_DEVICE=$(_file_device /localhost)
+else
+	export GEODESIC_LOCALHOST_MISSING=true
+fi
+
+function file_on_host() {
+  [[ $GEODESIC_LOCALHOST_MISSING != "true" ]] && [[  $(_file_device "$1") == ${GEODESIC_LOCALHOST_DEVICE} ]]
+}
+
+function _default_initial_wd() {
+	if [[ -d /stacks ]]; then
+		# Newer default using `atmos` and stacks
+		export GEODESIC_WORKDIR="/"
+	else
+		# Older default working directory
+		export GEODESIC_WORKDIR="/conf"
+	fi
+	red "# Defaulting initial working directory to \"${GEODESIC_WORKDIR}\""
+}
+
+# You can set GEODESIC_WORKDIR in your Geodesic preferences to have full control of your starting working directory
+if [[ -d $GEODESIC_WORKDIR ]]; then
+	[[ $SHLVL == 1 ]] && green "# Initial working directory configured as ${GEODESIC_WORKDIR}"
+else
+	if [[ -d $GEODESIC_HOST_CWD ]]; then
+		if [[ -n $LOCAL_HOME ]] && $(file_on_host "$GEODESIC_HOST_CWD"); then
+			export GEODESIC_WORKDIR=$(readlink -e "${GEODESIC_HOST_CWD}")
+			green "# Initial working directory set from host CWD to ${GEODESIC_WORKDIR}"
+		else
+			red "# Host CWD \"${GEODESIC_HOST_CWD}\" does not appear to be accessible from this container"
+			_default_initial_wd
+		fi
+	else
+		red "# No configured working directory is accessible:"
+		red "#    GEODESIC_WORKDIR is \"$GEODESIC_WORKDIR\""
+		red "#    GEODESIC_HOST_CWD is \"$GEODESIC_HOST_CWD\""
+		_default_initial_wd
+	fi
+fi
+
+[[ $SHLVL == 1 ]] && cd "${GEODESIC_WORKDIR}"
+
+unset -f _default_initial_wd
--- a/rootfs/etc/profile.d/atmos.sh
+++ b/rootfs/etc/profile.d/atmos.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+function _configure_atmos_base_path() {
+	# Leave $ATMOS_BASE_PATH alone if it is already set
+	if [[ -n $ATMOS_BASE_PATH ]]; then
+		if [[ $SHLVL == 1 ]]; then
+			green "# Using configured $ATMOS_BASE_PATH of \"$ATMOS_BASE_PATH\""
+		fi
+		return
+	fi
+
+	# If $GEODESIC_WORKDIR contains both a "stacks" and "components" directory,
+	# use it as the $ATMOS_BASE_PATH
+	if [[ -d "${GEODESIC_WORKDIR}/stacks" ]] && [[ -d "${GEODESIC_WORKDIR}/components" ]]; then
+		export ATMOS_BASE_PATH="${GEODESIC_WORKDIR}"
+		green "# Setting ATMOS_BASE_PATH to \"$ATMOS_BASE_PATH\" based on children of workdir"
+		return
+	fi
+
+	# If $GEODESIC_WORKDIR is a descendent of either a "stacks" or "components" directory,
+	# use the parent of that directory as ATMOS_BASE_PATH
+	if [[ "${GEODESIC_WORKDIR}" =~ /(stacks|components)/  ]]; then
+			if [[ "${GEODESIC_WORKDIR}" =~ /stacks/  ]]; then
+				export ATMOS_BASE_PATH="${GEODESIC_WORKDIR%/stacks/*}"
+			else
+				export ATMOS_BASE_PATH="${GEODESIC_WORKDIR%/components/*}"
+			fi
+		green "# Setting ATMOS_BASE_PATH to \"$ATMOS_BASE_PATH\" based on parent of workdir"
+		return
+	fi
+	yellow "# No candidate for ATMOS_BASE_PATH found, leaving it unset"
+}
+
+# Only configure ATMOS_BASE_PATH if we find an `atmos` executable,
+# but otherwise leave the function available for the user to call explicitly.
+# NOTE: If we start shipping `atmos` with Geodesic by default, change this to
+#   [[ -f /usr/local/etc/atmos/atmos.yaml ]] && _configure_atmos_base_path
+command -v atmos >/dev/null && _configure_atmos_base_path && unset -f _configure_atmos_base_path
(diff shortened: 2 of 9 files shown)


rootfs/etc/profile.d/atmos.sh
```diff
@@ -0,0 +1,38 @@
...
+			green "# Using configured $ATMOS_BASE_PATH of \"$ATMOS_BASE_PATH\""
+		fi
+		return
+	fi
+
+	# If $GEODESIC_WORKDIR contains both a "stacks" and "components" directory,
+	# use it as the $ATMOS_BASE_PATH
+	if [[ -d "${GEODESIC_WORKDIR}/stacks" ]] && [[ -d "${GEODESIC_WORKDIR}/components" ]]; then
+		export ATMOS_BASE_PATH="${GEODESIC_WORKDIR}"
+		green "# Setting ATMOS_BASE_PATH to \"$ATMOS_BASE_PATH\" based on children of workdir"
```
korenyoni, line 16: "Children of workdir" is a little vague outside of the context, especially to a Geodesic user who isn't reading the source of this script.

Here's a quick suggestion, but maybe word it differently.

```suggestion
		green "# Atmos base path automatically determined to be current workdir; setting ATMOS_BASE_PATH to \"$ATMOS_BASE_PATH\""
```

rootfs/etc/profile.d/atmos.sh
```diff
@@ -0,0 +1,38 @@
...
+
+	# If $GEODESIC_WORKDIR is a descendent of either a "stacks" or "components" directory,
+	# use the parent of that directory as ATMOS_BASE_PATH
+	if [[ "${GEODESIC_WORKDIR}" =~ /(stacks|components)/  ]]; then
+			if [[ "${GEODESIC_WORKDIR}" =~ /stacks/  ]]; then
+				export ATMOS_BASE_PATH="${GEODESIC_WORKDIR%/stacks/*}"
+			else
+				export ATMOS_BASE_PATH="${GEODESIC_WORKDIR%/components/*}"
+			fi
+		green "# Setting ATMOS_BASE_PATH to \"$ATMOS_BASE_PATH\" based on parent of workdir"
```
korenyoni, line 28:
```suggestion
		green "# Atmos base path automatically determined to be parent of current workdir; setting ATMOS_BASE_PATH to \"$ATMOS_BASE_PATH\""
```

rootfs/etc/profile.d/atmos.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+function _configure_atmos_base_path() {
+	# Leave $ATMOS_BASE_PATH alone if it is already set
+	if [[ -n $ATMOS_BASE_PATH ]]; then
+		if [[ $SHLVL == 1 ]]; then
```
korenyoni, line 6: Can't this be part of the same `if` conditional? i.e. `[[ -n $ATMOS_BASE_PATH ]] && [[ $SHLVL == 1 ]]` ? 

Also I think there should be an inline comment explaining why we are specifically testing for the number of shell levels.

korenyoni (approved):
I don't have any problems with the actual technicalities of this PR — to be honest there is a lot of shell parameter expansion syntax that is best understood by the author iteratively writing `_workdir.sh` and `atmos.sh`, so it's hard to nitpick anything there.

But I did leave a couple comments on what's printed out to the shell and also the grouping of if conditionals.

rootfs/etc/profile.d/atmos.sh
```diff
@@ -0,0 +1,38 @@
...
+	fi
+
+	# If $GEODESIC_WORKDIR is a descendent of either a "stacks" or "components" directory,
+	# use the parent of that directory as ATMOS_BASE_PATH
+	if [[ "${GEODESIC_WORKDIR}" =~ /(stacks|components)/  ]]; then
+			if [[ "${GEODESIC_WORKDIR}" =~ /stacks/  ]]; then
+				export ATMOS_BASE_PATH="${GEODESIC_WORKDIR%/stacks/*}"
+			else
+				export ATMOS_BASE_PATH="${GEODESIC_WORKDIR%/components/*}"
+			fi
```
korenyoni, line 27: 👍 

Re-reviewed shell parameter expansion syntax, LGTM

rootfs/etc/profile.d/atmos.sh
Nuru: What is printed out is in a form and style consistent with the other startup output from Geodesic:
```
# Mounting /Users/user into container
# Starting new geodesic session from cloudposse/geodesic:debian
# Exposing port 32380
# Setting ATMOS_BASE_PATH to "/localhost/source/geodesic" based on children of workdir
```

The fact that "children of workdir" is a bit cryptic is OK with me. The purpose of the message is not to _explain_ to the user how the setting was derived, as that just takes too much verbiage for a startup message. The point is to _distinguish which criteria were used_ in setting `ATMOS_BASE_PATH` so if there is a problem, we can more easily diagnose the cause.

rootfs/etc/profile.d/atmos.sh
Nuru: There is a classic example of a bad comment:
```
x = x + 1   // Add 1 to x
```
I didn't put a comment in about `SHLVL` because it seems to me it would be too much like that kind of bad comment. 

```bash
# Only output status in top level shell
if [[ $SHLVL == 1 ]]; then
```

The more useful comment would be about the bigger architectural design choice, but that would be long and does not belong in this bit of code. Since you asked, I will explain here in this comment.

The startup scripts (everything in `profile.d`) are responsible for a lot of the behavior of the interactive shell. Every time we spawn an interactive subshell, such as with `assume-role`, we run all these scripts again in the subshell to set it up. However, the startup messages are intended to expose configuration details of the Docker container or, if you will, the Geodesic session, more generally, and we do not want to repeat them when starting up a subshell with `assume-role` because (1) at that point they are not news and (2) its unnecessarily distracting. If you look around, you will find test for `$SHLVL == 1` all over the scripts in `profile.d`.

rootfs/templates/wrapper
```diff
@@ -156,20 +155,24 @@ function use() {
 	if [ "${local_home}" == "/localhost" ]; then
 		echo "WARNING: not mounting ${local_home} because it conflicts with geodesic"
 	else
-		echo "# Mounting ${local_home} into container"
-		DOCKER_ARGS+=(--volume=${local_home}:/localhost)
-		DOCKER_ARGS+=(--env LOCAL_HOME=${local_home})
+		echo "# Mounting ${local_home} into container with workdir ${GEODESIC_HOST_CWD}"
+		DOCKER_ARGS+=(
+		--volume="${local_home}:/localhost:"
```
nitrocode, line 160: Should there be a colon at the end of this string ?

```suggestion
		--volume="${local_home}:/localhost"
```

rootfs/etc/profile.d/_workdir.sh
```diff
@@ -0,0 +1,60 @@
+# Files in the profile.d directory are executed by the lexicographical order of their file names.
+# This file sets the working directory inside Geodesic to match the host directory Geodesic
+# was launched from, if possible. If the host directory is not accessible, it sets the working directory to `/`.
+#
+# This file is named _workdir.sh. The leading underscore is needed to ensure this file executes before
+# other files that may depend on it. The "w" is needed to ensure it is loaded *after* _preferences.sh
+#
+
+function _file_device() {
+	df --output=source "$1" | tail -1
```
nitrocode, line 10: nitpick: some functions are spaced with a single space and other functions have double space

nitrocode (approved):

osterman (dismissed):

rootfs/templates/wrapper
```diff
@@ -156,20 +155,24 @@ function use() {
...
+		--privileged
 		--publish ${GEODESIC_PORT}:${GEODESIC_PORT}
 		--name "${DOCKER_NAME}"
 		--rm
 		--env GEODESIC_PORT=${GEODESIC_PORT}
 		--env DOCKER_IMAGE="${DOCKER_IMAGE%:*}"
 		--env DOCKER_NAME="${DOCKER_NAME}"
 		--env DOCKER_TAG="${DOCKER_TAG}"
-		--workdir "${GEODESIC_WORKDIR}")
+		--env	GEODESIC_HOST_CWD="${GEODESIC_HOST_CWD}"
```
osterman, line 174:
```suggestion
		--env GEODESIC_HOST_CWD="${GEODESIC_HOST_CWD}"
```

rootfs/templates/wrapper
```diff
@@ -179,7 +182,7 @@ function use() {
 		if [ $# -eq 0 ]; then
 			set -- "/bin/bash" "-l" "$@"
 		fi
-		docker exec -it "${DOCKER_NAME}" $*
+		docker exec -it --env	GEODESIC_HOST_CWD="${GEODESIC_HOST_CWD}" "${DOCKER_NAME}" $*
```
osterman, line 185:
```suggestion
		docker exec -it --env GEODESIC_HOST_CWD="${GEODESIC_HOST_CWD}" "${DOCKER_NAME}" $*
```

nitrocode (approved):

korenyoni (dismissed):

osterman (approved):

PR #760, opened 2022-01-24
Configure atmos auto-completion, verify host filesystem detection

### LINKED ISSUE #757: AMD CPU got a warning for running Geodesic on Apple M1 CPU 🙂️
https://github.com/cloudposse/geodesic/blob/fb64a1d2655de0cbc39c5ab7976cf36ccd0dc68f/rootfs/etc/profile.d/banner.sh#L9

I think the first check should be `$(arch) == "x86_64" ` instead of `!=`

Nuru:
## what
- Notify user when Geodesic cannot distinguish host file system from container file system by device name and fall back to parsing file system path.
- Distinguish AMD CPU from Apple M1
- Configure `atmos` auto-completion when `atmos` is installed.

## why
- On Fedora, for example, the host and container filesystems are on the same device. This means we cannot tell if another host path was bind mounted or not. In this situation, we fall back to reporting that files under `/localhost` are on the HOST and other files are not, which was the test we were previously. (The newer device-based test is to support developers who keep their development file tree outside of their home file tree.)
- We warn users about trying to run Geodesic on Apple M1, but the warning was being displayed when running on genuine AMD CPUs, which is wrong. Closes #757.
- Convenience.

### DIFF
--- a/rootfs/etc/profile.d/_workdir.sh
+++ b/rootfs/etc/profile.d/_workdir.sh
@@ -14,10 +14,12 @@ function _file_device() {
 # Intended to support files on user-defined bind mounts in addition to `/localhost`.
 # This function is run by the command line prompt setup, so it should be very fast.
 # Therefore we cache some info in the environment.
-if df -a | grep -q /localhost; then
+if [[ $GEODESIC_LOCALHOST_DEVICE == "disabled" ]]; then
+	red "# Host filesystem device detection disabled."
+elif df -a | grep -q /localhost; then
 	export GEODESIC_LOCALHOST_DEVICE=$(_file_device /localhost)
 	if [[ $GEODESIC_LOCALHOST_DEVICE == $(_file_device /) ]]; then
-		red "# Host filesystem detection failed. Disabling HOST prompt."
+		red "# Host filesystem device detection failed. Falling back to \"path starts with /localhost\"."
 		GEODESIC_LOCALHOST_DEVICE="same-as-root"
 	fi
 else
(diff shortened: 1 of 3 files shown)


rootfs/etc/profile.d/_workdir.sh
```diff
@@ -16,12 +16,18 @@ function _file_device() {
...
+	fi
 else
-	export GEODESIC_LOCALHOST_MISSING=true
+	export GEODESIC_LOCALHOST_DEVICE="missing"
 fi
 
 function file_on_host() {
-	[[ $GEODESIC_LOCALHOST_MISSING != "true" ]] && [[ $(_file_device "$1") == ${GEODESIC_LOCALHOST_DEVICE} ]]
+	[[ $GEODESIC_LOCALHOST_DEVICE != "same-as-root" ]] && \
+	[[ $GEODESIC_LOCALHOST_DEVICE != "missing" ]] && \
```
leb4r, line 29: This also means that the `CWD` functionality is disabled as well (see conditional in `_default_initial_wd`). I'm personally okay with that because the default CWD works out of the box with `atmos` when things are copied into their corresponding directories.

leb4r:
LGTM, I left a small comment, but it can be fixed once a solution for the that one filesystem structure is determined. Everything else works as expected, including completion and the newly exposed  `atmos_configure_base_path`

leb4r (approved):

rootfs/etc/profile.d/_workdir.sh
Nuru: Fixed

leb4r (approved):
Both issues fixed! Old functionality is in place when the filesystem check fails. CWD working as expected, same with HOST prompt.

Issue #594, opened 2020-06-19
Files Written to Mounted Linux Home Directory Owned by Root User


osterman:
## what
* The user's shell inside Geodesic runs as `root`
* The script that launches Geodesic bind-mounts the host user's `$HOME` to `/localhost` to provide access to configuration files and allow for editing of host files
* Depending on the storage driver being used, it is possible that files created under `/localhost` from within Geodesic will be set to the same owner UID and GID (that is, owned by `root`) on the host as they have within Geodesic.
* Notably, this is _**not an issue**_ with Docker Desktop for Mac, which uses the `osxfs` storage driver to maintain a distinction between the UID and GID of a file as seen from within the Docker container versus what is set on the host file system. See [Ownership](https://web.archive.org/web/20190719101305/https://docs.docker.com/docker-for-mac/osxfs/#ownership) section of documentation.
* Update Feb 2022: this is _**not an issue**_ with Docker v20 on Ubuntu (tested on 20.04 LTS) when running in [rootless mode](https://docs.docker.com/engine/security/rootless/). As with macOS, this configuration correctly translates file ownership between the root UID and GID inside the container to the user's UID and GID on the host. 

## remediation discussion

Under MacOS, this is not a problem because the [OSXFS](https://docs.docker.com/docker-for-mac/osxfs/#ownership) driver handles UID and GID remapping automatically.

Under Linux, it is possible to use [User Namespaces](https://docs.docker.com/engine/security/userns-remap/) to remap the root user UID 0 to another UID (presumably the UID of the user running Geodesic), and do the same for GID, however, the documentation says you cannot use remapping when using the `--privileged` flag, which Geodesic requires, and even if you could, it would cause problems with `sudo` and setuid binaries in the container. 

#### further reading

- https://docs.docker.com/engine/security/userns-remap/
- https://blog.gougousis.net/file-permissions-the-painful-side-of-docker/
- https://seravo.fi/2019/align-user-ids-inside-and-outside-docker-with-subuser-mapping

Nuru:
@osterman @sboardwell 
I think we can use [BindFS](https://bindfs.org/) to solve this problem without risking the "works in my environment" problems that could come up if some users were running inside Geodesic as `root` and some were not. 

[...]

PR #762, opened 2022-01-30
Feat: Add `direnv allow` persistence between sessions

drmikecrowe:
## what

* Enables persistence of `direnv allow`  between sessions by dedicating `$HOME/.geodesic/direnv` to `/usr/share/xdg_data_home/direnv` in the container

## why

* `direnv allow` is not persisted between sessions

### DIFF
--- a/rootfs/etc/direnv/bash
+++ b/rootfs/etc/direnv/bash
@@ -7,10 +7,10 @@ unset BASH_ENV
 # Disable `direnv` output for scripts
 export DIRENV_LOG_FORMAT=
 
-# Allow current working directory (if we have permission)
-if [ -f .envrc ] && [ -w /etc/direnv/allow ]; then
-	direnv allow .
-fi
-
 # Process the `.envrc`
 eval "$(direnv export bash)"
+
+# Allow current working directory (if we have permission)
+if [ -f .envrc ] && [ -d /usr/share/xdg_data_home/direnv/allow ] && [ -w /usr/share/xdg_data_home/direnv/allow ]; then
+	direnv allow .
+fi
--- a/rootfs/etc/direnv/bash
+++ b/rootfs/etc/direnv/bash
@@ -11,6 +11,6 @@ export DIRENV_LOG_FORMAT=
 eval "$(direnv export bash)"
 
 # Allow current working directory (if we have permission)
-if [ -f .envrc ] && [ -d /usr/share/xdg_data_home/direnv/allow ] && [ -w /usr/share/xdg_data_home/direnv/allow ]; then
+if [ -f .envrc ] && [ -d /usr/share/xdg_data_home/direnv ] && [ -w /usr/share/xdg_data_home/direnv ]; then
 	direnv allow .
 fi
--- a/rootfs/etc/direnv/bash
+++ b/rootfs/etc/direnv/bash
@@ -7,10 +7,10 @@ unset BASH_ENV
 # Disable `direnv` output for scripts
 export DIRENV_LOG_FORMAT=
 
-# Process the `.envrc`
-eval "$(direnv export bash)"
-
 # Allow current working directory (if we have permission)
 if [ -f .envrc ] && [ -d /usr/share/xdg_data_home/direnv ] && [ -w /usr/share/xdg_data_home/direnv ]; then
 	direnv allow .
 fi
+
+# Process the `.envrc`
+eval "$(direnv export bash)"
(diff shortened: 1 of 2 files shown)


rootfs/templates/wrapper
```diff
@@ -109,6 +109,11 @@ function use() {
 		if [ -f "${geodesic_default_env_file}" ]; then
 			DOCKER_ARGS+=(--env-file=${geodesic_default_env_file})
 		fi
+		# persist direnv between sessions by putting in $HOME/.geodesic/direnv folder
+		local geodesic_folder="$(dirname "$geodesic_default_env_file")"
+		local geodesic_direnv_folder=${geodesic_folder}/direnv
+		mkdir -p $geodesic_direnv_folder
+		DOCKER_ARGS+=(--volume="$geodesic_direnv_folder:/usr/share/xdg_data_home/direnv")
```
korenyoni, line 116: https://github.com/cloudposse/geodesic/blob/ac1d4533be8e5ea8d4677ea29b12fca7c28461f2/os/alpine/Dockerfile.alpine#L62

```suggestion
		DOCKER_ARGS+=(--volume="$geodesic_direnv_folder:$XDG_DATA_HOME/direnv")
```

rootfs/etc/direnv/bash
```diff
@@ -8,7 +8,7 @@ unset BASH_ENV
 export DIRENV_LOG_FORMAT=
 
 # Allow current working directory (if we have permission)
-if [ -f .envrc ] && [ -w /etc/direnv/allow ]; then
+if [ -f .envrc ] && [ -d /usr/share/xdg_data_home/direnv ] && [ -w /usr/share/xdg_data_home/direnv ]; then
```
korenyoni, line 11:
```suggestion
if [ -f .envrc ] && [ -d "${XDG_DATA_HOME}/direnv" ] && [ -w "${XDG_DATA_HOME}/direnv" ]; then
```

https://github.com/cloudposse/geodesic/blob/ac1d4533be8e5ea8d4677ea29b12fca7c28461f2/os/alpine/Dockerfile.alpine#L62

korenyoni (commented):
LGTM but deferring to @Nuru because this makes the particular design decision to make use of `~/.geodesic/direnv` for `direnv allow` persistence.

korenyoni:
Edited PR description to match our [pull request format](https://github.com/cloudposse/geodesic/blob/8f987c7b3e7ec67449c886dd2a7412bb52d7e796/.github/PULL_REQUEST.md).

drmikecrowe:
My problem is solved if I set `DIRENV_ENABLED=true` in my Dockerfile.  

How should this be documented?  Had I know this, none of this would be necesary

korenyoni:
> My problem is solved if I set `DIRENV_ENABLED=true` in my Dockerfile.
> 
> How should this be documented? Had I know this, none of this would be necesary

I think we'll have to make a new place for it. At the very least a subsection in `customization.md`.

I've created https://github.com/cloudposse/geodesic/issues/768

drmikecrowe:
I can document this, as I want to add some customization docs when I submit my fix for #594 (which I need some more consensus before I submit)

Nuru (changes requested):
@drmikecrowe Thank you for this submission, and sorry for the confusion and the extra effort you had to go to find the solution.

We are not going to approve this PR for a bunch of reasons. Besides not being needed, if we wanted to persist the data on the host, we would want to have the option to do it on a per-image basis, for people that use more than one Geodesic-based Docker image. This would get pretty complicated, and it does not seem there is enough demand to justify the effort.

As for documentation (cc @korenyoni and #768), we currently document (to some extent) build-time configurations option in the Docker files. Specifically:
- [Dockerfile.custom](https://github.com/cloudposse/geodesic/blob/master/Dockerfile.custom)
- [Dockerfile.options](https://github.com/cloudposse/geodesic/blob/master/Dockerfile.options)
- [Dockerfile.debian](https://github.com/cloudposse/geodesic/blob/master/os/debian/Dockerfile.debian)
- [Dockerfile.alpine](https://github.com/cloudposse/geodesic/blob/master/os/alpine/Dockerfile.alpine)

Granted, it would be better to have the options more fully documented, and consolidated in one place, and easier to find. We will work on that: #768

For those wondering about runtime configuration, while we do have good[ documentation on how to _set_ runtime options](https://github.com/cloudposse/geodesic/blob/master/docs/customization.md), we unfortunately do not at the moment have much documentation on what the runtime options are. Your best resource for that at present is to review the code in the [profile.d](https://github.com/cloudposse/geodesic/tree/master/rootfs/etc/profile.d) directory and look for all uppercase environment variables. We try to make them reasonably self-documenting, but of course we acknowledge we could do a lot better. We do try to document them pretty well when we first introduce them, so you may find better information in the release notes. See, for example "**New options for customizing command line prompt appearance**" in the [Release Notes for v0.149.0](https://github.com/cloudposse/geodesic/releases/tag/0.149.0).

PR #763, opened 2022-01-30
Fix: show prompt even if you are not using `kube-ps1`

drmikecrowe:
## what

* Only attempt to print `kube_ps1` in prompt if you are using `kube-ps1`.

## why

* For non-kubernetes users, we sometimes encounter `kube_ps1 not found`.

### DIFF
--- a/rootfs/etc/profile.d/prompt.sh
+++ b/rootfs/etc/profile.d/prompt.sh
@@ -153,8 +153,13 @@ function geodesic_prompt() {
 		KUBE_PS1_KUBECONFIG_CACHE=""
 	fi
 
+	local kube_prompt=""
+	if typeset -f "${kube_ps1}" > /dev/null; then
+		kube_prompt="$(kube_ps1)"
+	fi
+
 	if [ -n "${BANNER}" ]; then
-		PS1=$' ${BANNER_MARK}'" ${BANNER} $(kube_ps1)${secrets_active}\n${tf_prompt}${dir_prompt}"
+		PS1=$' ${BANNER_MARK}'" ${BANNER} ${kube_prompt}${secrets_active}\n${tf_prompt}${dir_prompt}"
 	else
 		PS1="${tf_prompt}${dir_prompt}"
 	fi


korenyoni:
Hi @drmikecrowe 

Thanks again for the contribution. Again, next time please preserve the `what` and `why` sections 😃

rootfs/etc/profile.d/prompt.sh
```diff
@@ -168,9 +168,13 @@ function geodesic_prompt() {
...
 
+	local kube_prompt=""
+	if typeset -f "${kube_ps1}" > /dev/null; then
+		kube_prompt="$(kube_ps1)"
+	fi
+
 	if [ -n "${BANNER}" ]; then
-		# Intentional 2 spaces between banner mark and banner in order to offset it from level prompt
-		PS1=$' ${BANNER_MARK}'"  ${BANNER} $(kube_ps1)${secrets_active}\n${tf_prompt}${dir_prompt}"
+		PS1=$' ${BANNER_MARK}'"  ${BANNER} $(kube_prompt)${secrets_active}\n${tf_prompt}${dir_prompt}"
```
korenyoni, line 177: You are wrapping `kube_prompt` with braces (command substitution)

```bash
 ✗ . [none] (HOST) geodesic ⨠
bash: kube_prompt: command not found
```

```suggestion
		PS1=$' ${BANNER_MARK}'"  ${BANNER} ${kube_prompt}${secrets_active}\n${tf_prompt}${dir_prompt}"
```

rootfs/etc/profile.d/prompt.sh
```diff
@@ -168,9 +168,13 @@ function geodesic_prompt() {
 		KUBE_PS1_KUBECONFIG_CACHE=""
 	fi
 
+	local kube_prompt=""
+	if typeset -f "${kube_ps1}" > /dev/null; then
+		kube_prompt="$(kube_ps1)"
```
korenyoni, line 173:
```suggestion
		kube_prompt="${kube_ps1}"
```

korenyoni (changes requested):

korenyoni:
Can you please show clear steps on how you ran into this issue?

i.e. is the use-case if somehow `/etc/profile.d/promt:kube-ps1.sh` [is not added or is removed](https://github.com/cloudposse/geodesic/blob/ac1d4533be8e5ea8d4677ea29b12fca7c28461f2/os/alpine/Dockerfile.alpine#L152), or if the Dockerfile is modified to remove this step?

Also please confirm that it is not related to https://github.com/cloudposse/geodesic/issues/766

rootfs/etc/profile.d/prompt.sh
```diff
@@ -168,9 +168,13 @@ function geodesic_prompt() {
 		KUBE_PS1_KUBECONFIG_CACHE=""
 	fi
 
+	local kube_prompt=""
+	if typeset -f "${kube_ps1}" > /dev/null; then
```
korenyoni, line 172: We should be checking `typeset -f` against the function name, no need for interpolation. Also consider using `declare` instead of `typeset`.

```suggestion
	if typeset -f "kube_ps1" > /dev/null; then
```

drmikecrowe:
Closing since the core issue is actually #767 and that fix will fix this as well

PR #764, opened 2022-01-30
Feat: Better SSH Socket Support via `ssh-agent` (Linux Host)

drmikecrowe:
## what

This fixes firing up `ssh-agent` properly by testing two conditions:
* First, attempt to list added keys using `ssh-add -l &>/dev/null`.  If this returns an error of 2, we can't communicate with the agent
* Next, load the environment (if it exists) and repeat the above test.
If these fail, then restart the ssh-agent

Additionally, only map `SSH_AUTH_SOCK` from a real folder.  `/tmp` will not properly map into docker and results in being unable to access the socket.

Finally, if on Linux and using `SSH_AUTH_SOCK` _and_ you are not running as root (an upcoming PR), you have to map the host socket using `socat` to access the host folder as `root`. 

## why

* `ssh-agent` functionality is not robust enough (see above)


korenyoni:
Deferring to @Nuru because this affects high-level Geodesic wrapper functionality.

drmikecrowe:
@korenyoni -- I'm going to close this one as it's related to #594 and how that fix will impact the SSH socket

PR #765, opened 2022-01-30
Feat: Add Support for Linux Users and Custom Install Script

### LINKED ISSUE #594: Files Written to Mounted Linux Home Directory Owned by Root User

drmikecrowe:
* Support Linux non-root user -- fixes #594
* Add customization instructions to README
* Allow mapping of volumes in installation script

### DIFF
--- a/README.yaml
+++ b/README.yaml
@@ -27,7 +27,7 @@ screenshots:
   - name: "Demo"
     description: "<br/>[Example of running a shell](https://github.com/cloudposse/testing.cloudposse.co) based on the `cloudposse/geodesic` base docker image."
     url: "https://sweetops.com/wp-content/uploads/2019/03/termtosvg_fmnxoium.svg"
- 
+
 related:
   - name: "Packages"
     description: "Cloud Posse installer and distribution of native apps"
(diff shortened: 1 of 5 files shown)


korenyoni:
Hi @drmikecrowe, thanks for your first contribution!

Next time, please make use of `PULL_REQUEST_TEMPLATE`, i.e. preserving the _what_ and _why_ sections.

In terms of requesting changes or approval, I'm going to defer this to @Nuru

README.yaml
```diff
@@ -158,7 +158,40 @@ usage: |-
   Detailed information about launch-time configuration is in the [customization](./docs/customization.md)
   document, available from within the shell via `man customization`.
 
+  #### Customizing the installation script
```
korenyoni, line 161: Should this be in https://github.com/cloudposse/geodesic/blob/8f987c7b3e7ec67449c886dd2a7412bb52d7e796/docs/customization.md instead?

rootfs/templates/wrapper
```diff
@@ -187,7 +194,7 @@ function use() {
 		echo "# Starting new ${DOCKER_NAME} session from ${DOCKER_IMAGE}"
 		echo "# Exposing port ${GEODESIC_PORT}"
 		[ -z "${GEODESIC_DOCKER_EXTRA_ARGS}" ] || echo "# Launching with extra Docker args: ${GEODESIC_DOCKER_EXTRA_ARGS}"
-		docker run "${DOCKER_ARGS[@]}" ${GEODESIC_DOCKER_EXTRA_ARGS} ${DOCKER_IMAGE} -l $*
+		docker run "${DOCKER_ARGS[@]}" ${GEODESIC_DOCKER_EXTRA_ARGS} ${GEODESIC_ADDITIONAL_DOCKER_ARGS} ${DOCKER_IMAGE} -l $*
```
korenyoni, line 197: In terms of naming, it's hard to distinguish `GEODESIC_DOCKER_EXTRA_ARGS` and `GEODESIC_ADDITIONAL_DOCKER_ARGS`, they mean the same thing

rootfs/templates/wrapper
```diff
@@ -16,6 +16,8 @@ export GROUP_ID=$(id -g)
 export options=()
 export targets=()
 
+export GEODESIC_ADDITIONAL_DOCKER_ARGS="{{getenv "GEODESIC_ADDITIONAL_DOCKER_ARGS" ""}}"
```
korenyoni, line 19: See note below on naming

korenyoni (commented):
Some comments.

Again, I want to defer to @Nuru because he will likely suggest a different high level strategy for achieving this functionality.

drmikecrowe:
> Next time, please make use of `PULL_REQUEST_TEMPLATE`, i.e. preserving the _what_ and _why_ sections.
Let me resubmit, and pull apart the customization vs. fixing #594

Issue #719, opened 2021-06-27
Apple M1 (arm64) Support


dudymas:
See https://github.com/cloudposse/geodesic/issues/719#issuecomment-1005836850
----

per https://www.docker.com/blog/multi-platform-docker-builds/ , let's get geodesic on arm

This would allow apple silicon/m1 chipsets to run geodesic without emulation!

[![Slack Community](https://slack.cloudposse.com/badge.svg)](https://slack.cloudposse.com)

## Describe the Feature

Using multi-stage dockerfile features, introduce an arm build mode.

## Use Case

I have an m1 mac, and atm qemu struggles with several bugs on non-arm64 dockerfiles

## Describe Ideal Solution

Most tools should still be supported. There may be some that will have to wait while other folks update libraries or replace deprecated dependencies (i.e. terraform's hashicorp/template is archived and deprecated, and thus doesn't support darwin_arm64. It should be replaces with newer terraform templating providers.)

## Alternatives Considered

Running with qemu emulation hasn't worked. And many of the hacks people provide require substituting a stable qemu environment for an unstable and rather experimental one. I've found it to be even buggier, slower, and very inefficient.

## Additional Context

It's not expected to be a 100% solution, but if at least atmos is working, a lot of the rest of the tooling can follow soon!

Nuru:
@dudymas I understand the need, but simply compiling the Dockerfile for M1 will not be sufficient. Geodesic bundles dozens (hundreds?) of executable binaries and we would have to have M1 versions of all of them available to install. Plus we would need to build (compile and link) Python3 for M1, and right now we would have to do that with `qemu`, so if `qemu` is having problems, it likely won't work. 

I will leave this feature request open, but expect it will be more than a year before a sufficient infrastructure is in place to enable Geodesic to be delivered for `arm64`.

mikedizon:
has anyone tried running rocker using a Rosetta Terminal?

osterman:
# Apple M1 Support

We've confirmed that Geodesic works on recent versions of OSX on the M1 running as amd64 (not arm64). Docker auto-detects this by default, but otherwise, it’s possible to pass `--platform linux/amd64` to `docker` to force the platform.

nitrocode:
FYI to set the `--platform` or any additional docker argument, use the following environment variable.

```
export GEODESIC_DOCKER_EXTRA_ARGS="--platform linux/amd64"
```

PiotrPalkaSpotOn:
[...]

PR #771, opened 2022-02-15
Support Docker rootless mode & fix file ownership in root mode

### LINKED ISSUE #594: Files Written to Mounted Linux Home Directory Owned by Root User

### LINKED ISSUE #766: Entering Bash Subshell Results in Functions in `prompt.sh` and Prompt Hooks Not Being Found
Found a bug? Maybe our [Slack Community](https://slack.cloudposse.com) can help. 

[Slack Community](https://slack.cloudposse.com)

## Describe the Bug
Spawning an interactive subshell via `bash` results in a broken command line prompt and other errors.

**[Edit:** _yes, that is how Geodesic works. To spawn an interactive subshell, use_ `bash -l`**]**

## Steps to Reproduce
1. Enter Geodesic
2. Enter subshell via `bash`

## Expected Behavior

Functions in `prompt.sh` and prompt hooks should be properly sourced when entering a subshell, however bash complains about functions in `prompt.sh` and prompt hooks not being found (see below).

## Screenshots / Logs

(without a valid kubeconfig set via `KUBECONFIG`)

```bash
 ✗ . [none] (HOST) geodesic ⨠ 
 ⧉  geodesic 
 ✗ . [none] (HOST) geodesic ⨠ bash
bash: _kube_ps1_update_cache: command not found
bash: prompter: command not found
  geodesic 
```

(with a valid kubeconfig set via `KUBECONFIG`)

```bash
 ✗ . [none] (HOST) geodesic ⨠
 ⧉  geodesic (eg-mgmt-uw2-sandbox:default)
 ✗ . [none] (HOST) geodesic ⨠ bash
bash: _kube_ps1_update_cache: command not found
bash: prompter: command not found
   geodesic (eg-mgmt-uw2-sandbox:default)
```

## Environment (please complete the following information):

Anything that will help us triage the bug will help. Here are some ideas:
 - OS: OSX (host)
 - Geodesic Flavor: Alpine
 - Version `0.149.1`

## Additional Context

* Ran into this when attempting to reproduce #763

### LINKED ISSUE #767: Permissions of prompt:kube-ps1.sh causes issues with Linux operation as non-root user
## Describe the Bug
The permissions for `/etc/profile.d/prompt:kube-ps1.sh` as very restrictive.  This impacts #594.

## Steps to Reproduce
1. Execute `docker run -it --rm cloudposse/geodesic:latest-debian --login`
2. Execute:
```sh
 ⧉  geodesic 
 ✗ . [none] ~ ⨠ ls -l --color /etc/profile.d/prom*
-rwxrwxrwx 1 root root  7534 Jan 24 22:45 /etc/profile.d/prompt.sh
-rw------- 1 root root 10441 Jan  1  1970 /etc/profile.d/prompt:kube-ps1.sh
```

## Expected Behavior
All scripts should have similar permissions and not be restricted to just root

## Environment (please complete the following information):

Anything that will help us triage the bug will help. Here are some ideas:
 - OS: Linux Mint 
 - Version 20.2 (uma) -- based on Ubuntu (focal)
 - Geodesic version 0.149.1 based on Debian GNU/Linux 10 (buster) (10.11)

Nuru:
## what && why

- Install wrapper script in `$HOME/.local/bin` (per [XDG recommendation](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)) if `/usr/local/bin` is not writable. Supports preferred solution for #594, which is to have users run Docker in ["rootless"](https://docs.docker.com/engine/security/rootless/) mode.
- Provide option to use BindFS to map file owners on the host. Fixes #594 for users running `docker` daemon as root.
- Resolves #766 
- Resolves #767
- Downgrade the warning for Apple M1 Emulating Intel from seriously unsupported to a notice that there may be issues. Progress on #719 in that it appears the emulator is getting better, but still insufficient progress on native M1 support.
- Minor cleanups.

## notes

This release introduces a new customization: 
```
export GEODESIC_HOST_BINDFS_ENABLED=true
```
When Geodesic is launched by a shell where `$GEODESIC_HOST_BINDFS_ENABLED == true` it will configure `/localhost` so that files created by `root` inside Geodesic will have their ownership mapped to the same user ID and group ID as on the host. This a solution for people experiencing #594 and ONLY them. If your present solution is not creating files on the host owned by `root` then option will likely cause new problems. 

Note further that this is a workaround and not the desired solution. Our recommendation is that you resolve the underlying issue by [running Docker in "rootless" mode](https://docs.docker.com/engine/security/rootless/). This is much more secure and avoids the issue causing problems in the first place.


drmikecrowe:
By/Large, this works as expected and I'd approve this as-is.

I would suggest adding something about `GEODESIC_HOST_BINDFS_ENABLED` somewhere -- I had to search for how to enable this.

nitrocode (approved):

Dockerfile.custom
```diff
@@ -3,6 +3,12 @@
 # for a customer using the Cloud Posse Reference Architecture.
 # Use it as a basis for your own customizations.
 #
+# Note that Geodesic supports runtime customizations that
+# do not require a custome Dockerfile. See:
```
aknysh, line 7:
```suggestion
# do not require a custom Dockerfile. See:
```

PR #788, opened 2022-06-02
Preserve history in assume-role, prevent `make all` from attaching to old container

Nuru:
## what

* Preserve history in assume-role
* Prevent `make all` from attaching to old container

## why

* Better user experience if command history from top-level shell is available in subshell and vice versa
* Users get confused when `make all` succeeds but their shell does not include the latest changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-export DOCKER_IMAGE ?= cloudposse/geodesic
+export DOCKER_ORG ?= cloudposse
+export DOCKER_IMAGE ?= $(DOCKER_ORG)/geodesic
 export DOCKER_BASE_TAG ?= dev
 export DOCKER_BASE_OS ?= alpine
 export DOCKER_TAG ?= $(DOCKER_BASE_TAG)-$(DOCKER_BASE_OS)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -31,7 +33,7 @@ deps: init
 	docker tag $(DOCKER_IMAGE_NAME) $(DOCKER_IMAGE_NAME_BASE)
 
 %.install:
-	@docker run --rm --env DOCKER_IMAGE --env DOCKER_TAG $(DOCKER_IMAGE_NAME) | bash -s $(DOCKER_TAG)
+	@docker run --rm --env APP_NAME --env DOCKER_IMAGE --env DOCKER_TAG --env INSTALL_PATH $(DOCKER_IMAGE_NAME) | bash -s $(DOCKER_TAG)
```
nitrocode, line 36: Shouldn't this use `$()` for consistency with `$(DOCKER_IMAGE_NAME)` and `$(DOCKER_TAG)`

```suggestion
	@docker run --rm --env $(APP_NAME) --env $(DOCKER_IMAGE) --env $(DOCKER_TAG) --env $(INSTALL_PATH) $(DOCKER_IMAGE_NAME) | bash -s $(DOCKER_TAG)
```

rootfs/etc/profile.d/aws-okta.sh
```diff
@@ -63,7 +63,12 @@ if [ "${AWS_OKTA_ENABLED}" == "true" ]; then
 
 		shift
 		if [ $# -eq 0 ]; then
+			history -a # append history to file so it is available in subshell
 			aws-okta exec ${AWS_OKTA_ARGS[@]} $role -- bash -l
+			# read history from the subshell into the parent shell
+			# history -n does not work when HISTFILESIZE > HISTSIZE
+			history -c
```
nitrocode, line 70: Why clear the history first ?

nitrocode (approved):
Small nitpicks

rootfs/etc/profile.d/aws-okta.sh
Nuru: `history -r` will append the contents of the history file to the end of the history in memory. That will cause duplicates if the history is not cleared first.

Makefile
```diff
@@ -43,13 +45,26 @@ run:
 %.run: %.build %.install
 	@geodesic
 
+run/check:
+	@if [[ -n "$$(docker ps --format '{{ .Names }}' --filter name="^/$(APP_NAME)\$$")" ]]; then \
+		printf "**************************************************************************\n" ; \
+		printf "Not launching new container because old container is still running.\n"; \
+		printf "Exit all running container shells gracefully or kill the container with\n\n"; \
```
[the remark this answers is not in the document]
Nuru: Killing the containers does not exit them gracefully. Among other things, they will not save their history to the `$HISTFILE` when killed.

PR #795, opened 2022-07-07
feat: working towards supporting podman in wrapper script

Gowiem:
## what
* Updates `wrapper` to use `DOCKER_EXE` to work towards supporting podman (I don't claim that this works correctly yet)

## why
* Docker is becoming more and more bloatware and can be a huge pain. Podman and others are starting to gain traction and supporting them is of interest on my team.  

## references
* N/A

### DIFF
--- a/rootfs/templates/wrapper
+++ b/rootfs/templates/wrapper
@@ -197,18 +197,18 @@ function use() {
 
 	trap run_exit_hooks EXIT
 	# the extra curly braces around .ID are because this file goes through go template substitution local before being installed as a shell script
-	container_id=$(docker ps --quiet --filter name="^/${DOCKER_NAME}\$" --format '{{`{{ .ID }}`}}')
+	container_id=$(${DOCKER_EXE} ps --quiet --filter name="^/${DOCKER_NAME}\$" --format '{{`{{ .ID }}`}}')
 	if [ -n "$container_id" ]; then
 		echo "# Attaching to existing ${DOCKER_NAME} session ($container_id)"
 		if [ $# -eq 0 ]; then
 			set -- "/bin/bash" "-l" "$@"
 		fi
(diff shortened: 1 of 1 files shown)


rootfs/templates/wrapper
```diff
@@ -197,18 +197,18 @@ function use() {
```
nitrocode, line 197: Do you also need to update `rootfs/sbin/docker` ?

https://github.com/cloudposse/geodesic/blob/6bd1a5125cd32e9bcf71db0d10bda053dfbc4d69/rootfs/sbin/docker#L10

nitrocode:
@Gowiem friendly ping. I'd love to see podman support in geodesic. The docker app is very slow in comparison.

rootfs/templates/wrapper
nitrocode: or create a `rootfs/sbin/podman` ?

Gowiem: Probably... I'm assuming that is for DinD? That's likely low on my hit list for this PR, but is a good one to track for a future fix.

Gowiem:
@nitrocode yeah, I found that aliasing won't work as well. And you can build + run through podman, but I find a lot of the utility of Geodesic comes with the wrapper as it simplifies everything for non-Geodesic experts who need to run it. 

We have some Makefile work that utilizes this code that I believe got podman working, but it was painful and I (obviously) dropped it. If you want to chat through that, I'm happy to show it to you, but it'd likely be more confusing if I just shared a bunch of Makefile code here. 

I'll also add this draft PR to our open source board and we'll circle back around to cleaning this up at some point. I'm buying the new Macbooks once they launch for my FTE and myself -- we'll definitely come back around at that point because my ideal is that we don't even need to install Docker onto those machines at all 😁 🤞

PR #801, opened 2022-09-08
Update Debian, Alpine, etc, Preserve PATH

Nuru:
## what
- Copy Alpine `PATH` initialization to Debian, replacing Debian behavior of clobbering preexisting `PATH`
- Update Debian 11.3 -> 11.4
- Update Alpine 3.15.4 -> 3.16.2
- Miscellaneous cleanups

## why
- Debian behavior is to ignore any existing value of `PATH` environment variable when launching `bash` and set it to a predefined value. This makes it impossible to set the `PATH` in the Dockerfile. Alpine's approach (taken from Arch Linux) is to append the path components to any existing `PATH` value, allowing users to set a preferred path, but still ensuring that all expected/required directories are included in `PATH` to start.
- Keep up-to-date
- Clean up little things that were nevertheless irritating

## references
- Supersedes and closes #790 
- Supersedes and closes #799

### DIFF
--- a/os/alpine/Dockerfile.alpine
+++ b/os/alpine/Dockerfile.alpine
@@ -1,7 +1,7 @@
 # https://alpinelinux.org/
-ARG ALPINE_VERSION=3.15.4
+ARG ALPINE_VERSION=3.16.2
 # https://cloud.google.com/sdk/docs/release-notes
-ARG GOOGLE_CLOUD_SDK_VERSION=385.0.0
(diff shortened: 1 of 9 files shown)


joe-niland:
@Nuru I've tested this locally and it works well now when adding the following to the Dockerfile:

```
ENV PATH="/usr/local/go/bin:$PATH"
```

joe-niland (approved):

PR #805, opened 2022-09-28
Build arm64 native image on MacOS m1

PiotrPalkaSpotOn:
## what
* Dockerfile that you can actually build on Mac m1
* It's a Debian image (check "How to build" paragraph)
* since your cloudsmith docker repo doesn't have arm64 packages I used your [packages repo](https://github.com/cloudposse/packages) for packages installation that are not available in public repos
* looks like these packages are being installed in `/usr/local/bin` rather than `/usr/bin`
* the Dockerfile looks hacky at a glance. It definitely need some work to be pretty again. Populating the Debian repo with arm64 packages (and equivalent for Alpine) will definitely help.
* However, I didn't test all the functionality (just the bunch of commands I use everyday).
* geodesic prompt and `assume-role` works fine
* `/usr/local/bin/docs update` doesn't work during `docker build` phase (I described it in the Dockerfile)
* `/usr/bin/variant1` is missing.
* other things may be missing as well...

## why
* I made it out of curiosity if running the geodesic container in native (arm64) mode will be faster on my Mac m1 machine rather than keeping it in emulation mode
* I can't really say if it makes the difference as I just put it together and started using it.
* It's more like proof of concept for further development, however, it's usable

## references
* Looks like the discussion about this functionality is already in place https://github.com/cloudposse/geodesic/issues/719

## How to build:
```
sudo make -e DOCKER_BASE_OS=debian DOCKER_BUILD_FLAGS="--build-arg ARCH=arm64" build
```

### DIFF
--- a/os/debian/Dockerfile.debian
+++ b/os/debian/Dockerfile.debian
@@ -365,8 +365,14 @@ COPY docs/ /usr/share/docs/
 # Build man pages
 
 ## This throws error `#49 0.196 /bin/sh: 1: /usr/local/bin/docs: not found` on Mac m1 and I can't figure out why...
-RUN if [ "$(uname- m)" != 'aarch64']; then /usr/local/bin/docs update; fi
+RUN if [ "$ARCH" = "arm64" ]; then /usr/local/bin/docs update; fi
 
+# The `-slim` version of gomplate is not available for ARM 64 so we have to alter the DOWNLOAD_URL as the one in
+# `packages` repo is not compatible.
+RUN if [ "$ARCH" = "arm64" ] ; then \
+        cd packages && \
+        make -C install gomplate DOWNLOAD_URL='$(PACKAGE_REPO_URL)/releases/download/v$(PACKAGE_VERSION)/$(PACKAGE_NAME)_$(OS)-$(ARCH)' ; \
+    fi
 
 # Make sure that "user specific" directories we are sharing
 # are in fact available to all users
--- a/os/debian/Dockerfile.debian
+++ b/os/debian/Dockerfile.debian
@@ -11,6 +11,9 @@ ARG SESSION_MANAGER_PLUGIN_VERSION=latest
 
 ARG DEBIAN_FRONTEND=noninteractive
 
+# Multi platform support - override this arg with `arm64` string to build for Mac m1 for example
+ARG ARCH=amd64
+
 # Helm plugins:
 # https://github.com/databus23/helm-diff/releases
 ARG HELM_DIFF_VERSION=3.6.0
@@ -28,6 +31,10 @@ ARG PYTHON_VERSION=3.10.8
 # Debian comes with minimal Locale support. See https://github.com/docker-library/docs/pull/703/files
 # Recommended: LC_ALL=C.UTF-8
 ENV LC_ALL=C.UTF-8
+
+# Multi platform support - override this arg with `arm64` string to build for Mac m1 for example
+ARG ARCH=amd64
+ENV ARCH=${ARCH}
 
 # Once again, Debian is weird. Package-installed Python is system-only, and does not
 # play well with pip. See  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=765022#30
@@ -70,6 +77,10 @@ RUN find / -name __pycache__ -exec rm -rf {} \; -prune
 # Geodesic base image
 #
 FROM debian:$DEBIAN_VERSION
+
+# Multi platform support - override this arg with `arm64` string to build for Mac m1 for example
+ARG ARCH=amd64
+ENV ARCH=${ARCH}
 
 ARG VERSION
 ENV GEODESIC_VERSION=$VERSION
@@ -128,15 +139,15 @@ RUN curl -1sLf 'https://packages.cloud.google.com/apt/doc/apt-key.gpg' | APT_KEY
 
 # If runnig Dokcer build on Macos m1:
 # Since the packages are not available for arm64 let's build them using sources from Cloudposse's `package` repository
-RUN if [ "$(uname -m)" = 'aarch64' ] ; then \
+RUN if [ "$ARCH" = "arm64" ] ; then \
         # Install commonly known packages from public Debian repo
         apt-get update && apt-get install -y $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -E '/@(cloudposse|testing)/d' | sed -E 's/@community//g') && \
-        git clone https://github.com/SpotOnInc/packages.git -b fix-obsolete-packages && \
+        git clone https://github.com/cloudposse/packages.git && \
         cd packages && \
         # Packages with `@cloudposse` or `@testing` annotation have to be built from source
-        make -C install $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -E '/_amd64/d' | sed -nE 's/@(cloudposse|testing)//pg') ARCH=arm64 && \
+        make -C install $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -E '/_amd64/d' | sed -nE 's/@(cloudposse|testing)//pg') && \
         # # Packages ending with _amd64 have no arm64 equivalent
-        make -C install $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -nE 's/[@]?[a-zA-Z0-9]*_amd64//pg') ; \
+        make -C install $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -nE 's/[@]?[a-zA-Z0-9]*_amd64//pg') ARCH=amd64 ; \
     # On any other architecture proceed as usual. Packages will come from `https://dl.cloudsmith.io/public/cloudposse/packages/deb/debian bullseye main` repo.
     else \
         apt-get update && apt-get install -y \
@@ -171,7 +182,7 @@ RUN { gcloud config set core/disable_usage_reporting true --installation && \
 # RUN localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 
 # This directory doesn't exist on arm64 I guess because the next command throws an error "Directory doesn't exist" so let's make it. 
-RUN if [ "$(uname -m)" = 'aarch64' ] ; then mkdir -p /etc/sysctl.d ; fi
+RUN if [ "$ARCH" = "arm64" ] ; then mkdir -p /etc/sysctl.d ; fi
 
 RUN echo "net.ipv6.conf.all.disable_ipv6=0" > /etc/sysctl.d/00-ipv6.conf
 
@@ -303,12 +314,12 @@ RUN mv /usr/local/bin/aws /usr/local/bin/aws1 && \
 # ARG AWS_CLI_VERSION=2.2.8
 RUN AWSTMPDIR=$(mktemp -d -t aws-inst-XXXXXXXXXX) && \
     # We have to use platform-specific binary
-    if [ "$(uname -m)" = 'aarch64' ] ; then \
-        export ARCH='aarch64' ; \
+    if [ "$ARCH" = "arm64" ] ; then \
+        export AWS_ARCH='aarch64' ; \
     else \
-        export ARCH='x86_64' ; \
+        export AWS_ARCH='x86_64' ; \
     fi && \
-    curl -sSL "https://awscli.amazonaws.com/awscli-exe-linux-${ARCH}${AWS_CLI_VERSION:+-${AWS_CLI_VERSION}}.zip" -o "$AWSTMPDIR/awscliv2.zip" && \
+    curl -sSL "https://awscli.amazonaws.com/awscli-exe-linux-${AWS_ARCH}${AWS_CLI_VERSION:+-${AWS_CLI_VERSION}}.zip" -o "$AWSTMPDIR/awscliv2.zip" && \
     cd $AWSTMPDIR && \
     unzip -qq awscliv2.zip && \
     ./aws/install -i /usr/share/aws/v2 -b /usr/share/aws/v2/bin && \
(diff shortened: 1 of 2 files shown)


PiotrPalkaSpotOn:
Happy to get your feedback how it can be improved or any other ideas if we can work further on this functionality.

Nuru:
Thank you very much for this effort.

The key question is: how much of a performance improvement does this give you. It will be a big effort to support fully (see https://github.com/cloudposse/packages/issues/2872#issuecomment-1272259577), and if it does not make a big difference to users, it is less of a priority to do it.

Nuru:
Why did you remove `stern`? The project was forked, but we expect to keep using the forked version.

PiotrPalkaSpotOn:
@Nuru I think stern package disappeared from `package` repo the time I was building it thus removed from the list. Let me double check that

PiotrPalkaSpotOn:
From what I see so far there's not much performance difference on daily usage on my mac m1 so far. I want to announce among my colleagues how they can build mac m1 image. Maybe I'll get some feedback and see what broader community in my company think about this.

PiotrPalkaSpotOn:
[FZF package upgrade required](https://github.com/cloudposse/packages/pull/3189) in `packages` repository as the old one doesn't support arm64 architecture. I didn't tested it, tho, as I don't know the use case.

packages.txt
```diff
@@ -1,10 +1,10 @@
 # Essential alpine packages
-awless@cloudposse
+awless@cloudposse_amd64
```
PiotrPalkaSpotOn, line 2: Some maintainers don't support arm64 packages in their releases. I had to filter them out and use x86 packages instead.

PiotrPalkaSpotOn:
Ok, I found there's load of performance improvement. From subjective standpoint I can feel `atmos terraform plan` goes a bit faster which is productivity improvement when I execute this command dozens of times daily

dudymas:
So you mention in the description running make with `sudo`. So far it looks like I don't need to do that on my m1 mac. Is there anything I might have missed?

dudymas:
aha, so I checked if I run the command, it looks like it only builds amd64. I tried with sudo, and still amd64. If I try the build-arm64 it wants a custom image, which unfortunately won't work for us. The base image should be a public image rather than a private ECR image.

PiotrPalkaSpotOn:
@dudymas seems like over the course of development I added one extra variable `ARCH` but I didn't update the build instruction. It should be 

```bash
sudo make -e DOCKER_BASE_OS=debian ARCH=arm64 build
```

thanks for having a look.

dudymas:
okay, I got further! the builds are nontheless unable to install atmos/terraform yet. I'm digging around and will try to make some suggestions soon.

PiotrPalkaSpotOn:
@dudymas if you're willing to spend some time on that I'm eager to help and try to solve the problems so the build is reproducible on your machine. Let me know if git comments section is the right place to do so or should we have a separate slack channel

kevcube:
@PiotrPalkaSpotOn see [here](https://docs.docker.com/build/building/multi-platform/) there are default args for BUILDARCH and TARGETARCH so you should not need to provide them in --build-args

os/debian/Dockerfile.debian
```diff
@@ -9,6 +9,11 @@ ARG KUBE_PS1_VERSION=0.8.0
 # https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html#plugin-version-history
 ARG SESSION_MANAGER_PLUGIN_VERSION=latest
 
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Multi platform support - override this arg with `arm64` string to build for Mac m1 for example
+ARG ARCH=amd64
```
kevcube, line 15:
```suggestion
# Multi platform support - This value is set by docker when using `--platform=linux/amd64|linux/arm64`
ARG TARGETARCH
```

os/debian/Dockerfile.debian
```diff
@@ -27,6 +32,10 @@ ARG PYTHON_VERSION=3.10.8
 # Recommended: LC_ALL=C.UTF-8
 ENV LC_ALL=C.UTF-8
 
+# Multi platform support - override this arg with `arm64` string to build for Mac m1 for example
+ARG ARCH=amd64
+ENV ARCH=${ARCH}
+
```
kevcube, line 38: seems this is a dupe of line 15
```suggestion
```

os/debian/Dockerfile.debian
```diff
@@ -72,6 +78,10 @@ RUN find / -name __pycache__ -exec rm -rf {} \; -prune
 #
 FROM debian:$DEBIAN_VERSION
 
+# Multi platform support - override this arg with `arm64` string to build for Mac m1 for example
+ARG ARCH=amd64
+ENV ARCH=${ARCH}
+
```
kevcube, line 84: again a duplicate
```suggestion
```

os/debian/Dockerfile.debian
```diff
@@ -127,9 +137,27 @@ RUN curl -1sLf 'https://dl.cloudsmith.io/public/cloudposse/packages/cfg/setup/ba
 RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
 RUN curl -1sLf 'https://packages.cloud.google.com/apt/doc/apt-key.gpg' | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=true apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
 
-RUN apt-get update && apt-get install -y \
-    $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -E 's/@(cloudposse|community|testing)//g' ) && \
-    mkdir -p /etc/bash_completion.d/ /etc/profile.d/ /conf && \
+# If runnig Dokcer build on Macos m1:
```
kevcube, line 140:
```suggestion
# If running Docker build on arm64:
```

os/debian/Dockerfile.debian
```diff
@@ -127,9 +137,27 @@ RUN curl -1sLf 'https://dl.cloudsmith.io/public/cloudposse/packages/cfg/setup/ba
...
+        # The `-slim` version of gomplate is not available for ARM 64 so we have to alter the DOWNLOAD_URL as the one in
+        # `packages` repo is not compatible.
+        make -C install gomplate DOWNLOAD_URL='$(PACKAGE_REPO_URL)/releases/download/v$(PACKAGE_VERSION)/$(PACKAGE_NAME)_$(OS)-$(ARCH)' ; \
+    # On any other architecture proceed as usual. Packages will come from `https://dl.cloudsmith.io/public/cloudposse/packages/deb/debian bullseye main` repo.
+    else \
+        apt-get update && apt-get install -y \
+            $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -E 's/@(cloudposse|community|testing).*//g' ) ; \
+    fi
+
+RUN mkdir -p /etc/bash_completion.d/ /etc/profile.d/ /conf && \
```
kevcube, line 160:
```suggestion
# Since the packages are not available for arm64 let's build them using sources from Cloudposse's `package` repository
RUN if [ "${TARGETARCH}" = "arm64" ] ; then \
        # Install commonly known packages from public Debian repo
        apt-get update && apt-get install -y $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -E '/@(cloudposse|testing)/d' | sed -E 's/@community//g') && \
        git clone https://github.com/cloudposse/packages.git && \
        cd packages && \
        # Packages with `@cloudposse` or `@testing` annotation have to be built from source
        make -C install $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -E '/_amd64/d' | sed -nE 's/@(cloudposse|testing)//pg') && \
        # # Packages ending with _amd64 have no arm64 equivalent
        make -C install $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -nE 's/[@]?[a-zA-Z0-9]*_amd64//pg') ARCH=amd64 && \
        # The `-slim` version of gomplate is not available for arm64 so we have to alter the DOWNLOAD_URL as the one in
        # `packages` repo is not compatible.
        make -C install gomplate DOWNLOAD_URL='$(PACKAGE_REPO_URL)/releases/download/v$(PACKAGE_VERSION)/$(PACKAGE_NAME)_$(OS)-$(TARGETARCH)' ; \
    # On any other architecture proceed as usual. Packages will come from `https://dl.cloudsmith.io/public/cloudposse/packages/deb/debian bullseye main` repo.
    else \
        apt-get update && apt-get install -y \
            $(grep -h -v '^#' /etc/apt/packages.txt /etc/apt/packages-debian.txt | sed -E 's/@(cloudposse|community|testing).*//g' ) ; \
    fi

RUN mkdir -p /etc/bash_completion.d/ /etc/profile.d/ /conf && \
```

os/debian/Dockerfile.debian
```diff
@@ -156,6 +184,9 @@ RUN { gcloud config set core/disable_usage_reporting true --installation && \
 # "locales" package and run the following command:
 # RUN localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 
+# This directory doesn't exist on arm64 I guess because the next command throws an error "Directory doesn't exist" so let's make it. 
+RUN if [ "$ARCH" = "arm64" ] ; then mkdir -p /etc/sysctl.d ; fi
```
kevcube, line 188:
```suggestion
RUN if [ "$TARGETARCH" = "arm64" ] ; then mkdir -p /etc/sysctl.d ; fi
```

os/debian/Dockerfile.debian
```diff
@@ -285,7 +316,13 @@ RUN mv /usr/local/bin/aws /usr/local/bin/aws1 && \
 # but it is updated several times a week, so we choose to just get the latest.
 # ARG AWS_CLI_VERSION=2.2.8
 RUN AWSTMPDIR=$(mktemp -d -t aws-inst-XXXXXXXXXX) && \
-    curl -sSL "https://awscli.amazonaws.com/awscli-exe-linux-x86_64${AWS_CLI_VERSION:+-${AWS_CLI_VERSION}}.zip" -o "$AWSTMPDIR/awscliv2.zip" && \
+    # We have to use platform-specific binary
+    if [ "$ARCH" = "arm64" ] ; then \
```
kevcube, line 320:
```suggestion
    if [ "$TARGETARCH" = "arm64" ] ; then \
```

os/debian/Dockerfile.debian
```diff
@@ -306,16 +343,32 @@ ENV AWS_PAGER="less -FRX"
 
 # Install AWS Session Manager Plugin
 ARG SESSION_MANAGER_PLUGIN_VERSION
-RUN curl -L "https://s3.amazonaws.com/session-manager-downloads/plugin/${SESSION_MANAGER_PLUGIN_VERSION}/ubuntu_64bit/session-manager-plugin.deb" \
+RUN if [ "$ARCH" = "arm64" ] ; then \
```
kevcube, line 346:
```suggestion
RUN if [ "$TARGETARCH" = "arm64" ] ; then \
```

os/debian/Dockerfile.debian
```diff
@@ -306,16 +343,32 @@ ENV AWS_PAGER="less -FRX"
...
+        export  SESSION_MANAGER_PLUGIN_ARCH='arm64' ; \
+    else \
+        export SESSION_MANAGER_PLUGIN_ARCH='64bit' ; \
+    fi && \
+    curl -L "https://s3.amazonaws.com/session-manager-downloads/plugin/${SESSION_MANAGER_PLUGIN_VERSION}/ubuntu_${SESSION_MANAGER_PLUGIN_ARCH}/session-manager-plugin.deb" \
```
kevcube, line 351: This is ugly, thanks Amazon 🙃

os/debian/Dockerfile.debian
```diff
@@ -306,16 +343,32 @@ ENV AWS_PAGER="less -FRX"
...
+#         INSTALL_PATH=/usr/bin make -C install variant && \
+#         mv /usr/bin/variant /usr/bin/variant1 ; \
+#     fi
+
+
 # Build man pages
-RUN /usr/local/bin/docs update
+
+## This throws error `#49 0.196 /bin/sh: 1: /usr/local/bin/docs: not found` on Mac m1 and I can't figure out why...
+RUN if [ "$ARCH" != "arm64" ]; then /usr/local/bin/docs update; fi
```
kevcube, line 371:
```suggestion
RUN if [ "$TARGETARCH" != "arm64" ]; then /usr/local/bin/docs update; fi
```

kevcube (commented):
according to [this blog post](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/), here is the complete list

```
BUILDPLATFORM — matches the current machine. (e.g. linux/amd64)

BUILDOS — os component of BUILDPLATFORM, e.g. linux

BUILDARCH — e.g. amd64, arm64, riscv64

BUILDVARIANT — used to set ARM variant, e.g. v7

TARGETPLATFORM — The value set with --platform flag on build

TARGETOS - OS component from --platform, e.g. linux

TARGETARCH - Architecture from --platform, e.g. arm64

TARGETVARIANT
```

os/debian/Dockerfile.debian
PiotrPalkaSpotOn: nope, different image. https://github.com/cloudposse/geodesic/blob/4851de6c2419149a0c3500955816956c3e17c9ee/os/debian/Dockerfile.debian#L27 did you have a chance to make changes locally and build it?

os/debian/Dockerfile.debian
PiotrPalkaSpotOn: Again, different image.

PiotrPalkaSpotOn:
@kevcube thanks for input and note about `TARGETARCH` usage! :) I'll give it a try locally and update the PR. I also left a comment about "duplicated" `ARCH` appearances that are indeed duplicated in the file but not in given context

os/debian/Dockerfile.debian
PiotrPalkaSpotOn: I think `TARGETARCH` will straight the code and some of this won't be necessary. I think I'll go with `BUILDARCH`, tho, as it's `TARGETARCH` appears to be available only when using `--platform` parameter so the latter one seems more stable. Correct me if I'm wrong

kevcube: Oops I didn't read the whole file, only the snippets in the diff, didn't realize it is a multi-stage build

IIRC TARGETARCH and BUILDARCH are both available, and BUILDARCH is the arch of the host machine so it won't work for doing a multi-platform build in CI

But again this is from memory 😝 I will test today and let you know

kevcube: TARGETARCH should be set on all buildkit-based builds which is default in recent Docker but I'm seeing it must be placed after the FROM. I will make a PR to your fork with a working build.

kevcube:
@PiotrPalkaSpotOn see https://github.com/SpotOnInc/geodesic/pull/1

Tested and running

kevcube:
I want this PR to also include Alpine so I am working on that.

I think ultimately it is best that we update the cloudposse packages repo where we can to also build for arm64 and then we can avoid the hack of us building those packages in container build. But that is a long term effort that is not high priority to me

PR #806, opened 2022-09-29
add - to options for yq

fcooper-spoton:
## what
* add required '-' to options for the yq command

## why
* script fails to run without them


Nuru:
@fcooper-spoton Thank you for this contribution.

Unfortunately, the interface for `yq` changed significantly between now and the time these scripts were written. Adding a `-` to the `r` may cause the commands to not error out, but does not cause them to work as intended. 

We will fix our scripts in a future PR.

PR #825, opened 2022-12-19
ECR Builds

Benbentwo:
## what
* Adds ECR Builds

## why
* Occasionally get dockerhub rate limited, public ECR fixes that

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -13,12 +13,28 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    env:
+      IAM_ROLE_SESSION_NAME: cloudposse/geodesic/ci
+      AWS_REGION: us-east-1
+
     strategy:
       matrix:
         os: ['alpine', 'debian']
     steps:
     - name: "Checkout source code at current commit"
       uses: actions/checkout@v2
+
+    - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-region: ${{ env.AWS_REGION }}
+        role-to-assume: ${{ env.ECR_AWS_ROLE }}
+        role-session-name: ${{ env.IAM_ROLE_SESSION_NAME }}
+
+    - name: Login to Public Amazon ECR
+      run: |
+        aws ecr-public get-login-password --region ${{ env.AWS_REGION }} | docker login --username AWS --password-stdin public.ecr.aws
+
     - name: Prepare tags for Docker image
       id: prepare
       # Even though we are no longer releasing plain numbered version tags like "0.111.0"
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -10,12 +10,31 @@ on:
 jobs:
   build-and-push:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    env:
+      IAM_ROLE_SESSION_NAME: geodesic-ci
+      AWS_REGION: us-east-1
+      ECR_REGISTRY: public.ecr.aws/h1u4b2i4/
```
osterman, line 19: Is this customizable?

Benbentwo: Yes but needs to get approved by amazon, put in request a while ago, might need some verification

.github/workflows/docker.yml
```diff
@@ -10,12 +10,32 @@ on:
 jobs:
   build-and-push:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    env:
+      IAM_ROLE_SESSION_NAME: geodesic-ci
+      AWS_REGION: us-east-1
+      ECR_REGISTRY: public.ecr.aws/ # This will become `public.ecr.aws/cloudposse/$repositoryName`
```
osterman, line 19:
```suggestion
      ECR_REGISTRY: public.ecr.aws/ # Images will be published to `public.ecr.aws/cloudposse/$repositoryName`
```

osterman (approved):

.github/workflows/docker.yml
```diff
@@ -10,12 +10,32 @@ on:
...
+      with:
+        role-to-assume: ${{ secrets.ECR_AWS_ROLE }}
+        aws-region: ${{ env.AWS_REGION }}
+        role-session-name: ${{ env.IAM_ROLE_SESSION_NAME }}
+
+    - name: Login to Public Amazon ECR
+      run: |
+        docker logout public.ecr.aws
+        aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws
+
```
nitrocode, line 38: You could also add the metadata action

```suggestion

      - name: Docker meta
        id: meta
        uses: docker/metadata-action@v4
        with:
          images: cloudposse/geodesic
```

Benbentwo: I propose we make this update in a future PR similar to @osterman's comment below

nitrocode: sure but the tagging is going to be much more complicated than adding the labels. I would recommend 2 separate PRs. 

1. PR to add labels and metadata action
2. PR to modify the complicated tags logic below to use metadata output

PR #837, opened 2023-03-16
Initial support for ARM

### LINKED ISSUE #719: Apple M1 (arm64) Support

Nuru:
## Breaking changes

In addition to the changes listed below, this release may bring unexpected breaking changes, ironically due to support for ARM.

Previously, if you built a Docker image based on Geodesic on an ARM machine like an Apple M1, because Geodesic was only available in `linux/amd64` architecture, your build would have been forced into `linux/amd64` architecture and your installations, whether from `apt-get` or other sources, would have had to have been the same `linux/amd64` architecture to work. (The build and run of the resulting Docker image would have been run under emulation.) Now, if you are building on an Apple M1 or M2, you will get the Geodesic `linux/arm64` architecture (using native, not emulated code for build and execution, with potentially huge performance benefits), and the rest of your Dockerfile will need to be updated to install architecture-specific packages, some of which may not exist. You will need to decide if you want to go on without them or rather stick to `linux/amd64` emulation to retain them.

Furthermore, if you built and pushed a Geodesic image in the past, you would always get a single architecture (`linux/amd64`) image. Now, if you are not careful, you may overwrite that with a `linux/arm64` image, causing slowdowns for people using your image on Intel/AMD/X86_64 hardware. If you want to support both architectures in a shared image, you will need to use Docker `buildx` to generate a multi-platform image. 

If you want to avoid all this, you can convert your `FROM` statement in your Dockerfile from
```
FROM cloudposse/geodesic:2.0.0-debian
```
to
```
FROM --platform=linux/amd64 cloudposse/geodesic:2.0.0-debian
```

For more information on multi-platform (a.k.a. multi-architecture) builds, see:
- [Multi-platform images](https://docs.docker.com/build/building/multi-platform/)
- [Faster multi-platform builds](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/)

 
## what
- Enhance Debian version of Geodesic to run on `arm64` as well as `amd64` architecture
- Make `cloudposse/geodesic:latest` image Debian instead of Alpine
- Remove binaries that are not available on `arm64` or are just outdated:
   - `kops`
   - `awless`
   - `cfssl`
   - `rakkess`
   - `tfenv`
   - `tfmask`
- Remove `init-terraform` script meant to be used with obsolete `direnv`/`tfenv` configuration
- Copy `python` from "official" pre-built docker images rather than compiling it ourselves
- Update Python 3.10.8 -> 3.10.10
- Update Google Cloud SDK 410.0.0 -> 422.0.0 ([breaking changes](https://cloud.google.com/sdk/docs/release-notes#42200_2023-03-14))
- Update Helm-git 0.14.0 -> 0.15.1

## why
* Provide native code support for Apple and Gravitron hardware

## notes

This is our initial support of `arm64` and can be expected to have some bugs to shake out.

We are only supporting `arm64` on Debian at this time. We will not support it on Alpine. Will consider supporting CentOS (or its successor) if we have sufficient demand. 

Geodesic relies heavily on Cloud Posse's [packages](https://github.com/cloudposse/packages/) distribution, and it has not yet been updated to automatically generate `arm64` packages. As a result, for most packages, only the latest version is available in `arm64`. We have historical versions of `atmos`, `kubectl`, and `terraform` published. If you need historical versions of other packages, you can request them by opening an issue in [packages](https://github.com/cloudposse/packages/), but please consider either staying on `amd64` or updating to the latest version of the binary instead. Please also give us a few weeks to get `arm64` packages automated.

## references
- Supersedes and closes #805 
- Closes #719


kevcube:
Cool! Unfortunate that Cloudposse is not migrating Alpine. Why is that? I prefer the alpine package repos because they usually have much more recent versions of software

PR #841, opened 2023-04-03
Install stubs for tools missing arm64 support

Nuru:
## what
- For tools we have been including in Geodesic but that do not have `arm64` support:
   - Install the tool in the `amd4` versions of Geodesic
   - Install a stub script that explains the tool is missing due to lack of support

## why
* In #837, we removed all the tools without `arm64` support from both versions. However, those tools are used by scripts and commands we continue to include, and we do not know how important those are to the Geodesic community. This way, they will continue to work as before for `amd64` versions and will generate meaningful error messages (instead of generic "command not found" errors) in the `arm64` version.

### DIFF
--- a/packages-amd64-only.txt
+++ b/packages-amd64-only.txt
@@ -1,0 +1,9 @@
+# These are packages that are only supported on amd64,
+# typically because they have not been updated since 2021.
+awless@cloudposse
+cfssl@cloudposse
+emailcli@cloudposse
+goofys@cloudposse
+rakkess@cloudposse
+tfenv@cloudposse
+tfmask@cloudposse
--- a/rootfs/usr/local/bin/no-arm64-support
+++ b/rootfs/usr/local/bin/no-arm64-support
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+printf "$(tput setaf 1)%s is not supported on this platform (arm64)$(tput setaf 0)\n" "$(basename "$0")" >&2
+exit 1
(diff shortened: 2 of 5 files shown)


rootfs/usr/local/bin/no-arm64-support
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+printf "$(tput setaf 1)%s is not supported on this platform (arm64)$(tput setaf 0)\n" "$(basename "$0")" >&2
```
osterman, line 3: Nice.

packages-amd64-only.txt
```diff
@@ -0,0 +1,9 @@
+# These are packages that are only supported on amd64,
+# typically because they have not been updated since 2021.
+awless@cloudposse
+cfssl@cloudposse
+emailcli@cloudposse
+goofys@cloudposse
+rakkess@cloudposse
+tfenv@cloudposse
+tfmask@cloudposse
```
osterman, line 9: I do believe we should deprecate 💯 of these from our base image. We can do that in a follow on PR. None of these are critical to have in the base image, and it's easy enough for users to mitigate it in their downstream images. Since 2.x has many breaking changes, this would be the time, no?

osterman (dismissed):

packages-amd64-only.txt
Nuru: @osterman The difficult one to remove is `goofys`, because we have commands `s3` and `s3fs` (that work together to mount an S3 bucket as if it were a local disk) that depend on it. That was my motivation to restore these commands and add stubs. Do we want to remove `s3` and `s3fs` as well?

PR #898, opened 2023-12-19
Update Demo GIF with VHS

osterman:
## what
* Updated demo with automation
* See demo

## why
* Video was hopelessly out of date, with this change we can keep it regularly updated with each PR.


demo.tape
```diff
@@ -0,0 +1,64 @@
...
+Set Padding 12
+
+Set Width 1200
+Set Height 600
+
+Set PlaybackSpeed 1
+
+Set Shell bash
+
+Type "# After having run `make all`, simply run `geodesic` to enter the interactive shell" Enter
```
Benbentwo, line 33: IMO make run is more commonly used, as people may customize the startup script name, but make run is always the same

```suggestion
Type "# After having run `make all`, simply run `make run` (or `geodesic`) to enter the interactive shell" Enter
```

Benbentwo (dismissed):
nitpick

Benbentwo (approved):

PR #940, opened 2024-05-10
PR & Release management maintenance

Nuru:
## what

- Install font needed by `vhs-action` in a separate step
- Label PRs that only affect the Alpine version as `alpine-only`
- Exclude Apline-only PRs from release notes
- Add a standard catchall to every release note

## why

- Work around broken `install-fonts` in `vhs-action`
- Deprecate Alpine, focus release notes on Debian
- The AWS CLI v1 and `boto3` release notes are huge and yet not very informative
- Packages are updated on every release, without those changes being documented, so let people know that

## references

- https://github.com/charmbracelet/vhs-action/issues/264


demo.tape
```diff
@@ -35,23 +35,18 @@ Sleep 500ms
...
-Type "# It ships with common commands like `terraform`" Enter
-Sleep 500ms
-
-Type "terraform -help" Enter
-Sleep 500ms
-
-Type "# ...and the `aws` CLI =)" Enter
-Sleep 500ms
-
-Type "aws help" Enter
```
osterman, line 49: What can we showcase then that make it more than `jq`. Not that it was awe-inspiring in the previous version, this is definitely less.

osterman: Maybe we at least bundle `atmos`

Nuru: IDK. I took out Terraform in favor of having the option to use Tofu, and `aws` in preparation for the cloud-agnostic version of Geodesic. Really `kubectl` is the biggest thing left, and after that, it is mostly the scripts, prompt, integration, etc. that make Geodesic wonderful. It is great that we have a robust set of tools, but they mostly follow the Unix preference of doing one thing well and being composable, so not flashy on their own.

I suppose we could `ls /usr/local/bin`, which is empty on `debian-slim`.

```
2to3@		      docs*			    pip3.12*
2to3-3.12*	      eks-update-kubeconfig*	    pydoc@
annotate-namespace*   eks-update-kubeconfig.draft*  pydoc3@
args*		      emailcli@			    pydoc3.12*
awless@		      geodesic-aptindex-md5*	    python@
aws@		      git-credential-github*	    python-config@
aws-config-setup*     goofys@			    python3@
aws-region*	      grafana-db*		    python3-config@
aws2@		      helm3-metadata-fix*	    python3.12*
aws2_completer@       helmctl*			    python3.12-config*
aws_completer@	      idle@			    s3*
boot*		      idle3@			    session-manager-plugin@
build-kops-manifest*  idle3.12*			    short.txt
build-kubecfg*	      init*			    terminal-server
cfssl@		      install-docker-command*	    tfenv@
codefresh-pipeline*   kopsctl*			    tfmask@
codefresh-settings*   kubectl-auto-select*	    tmate-session*
conf-directory*       make*			    to_short.txt
confirm*	      new-project*		    twiddle*
crudini*	      no-arm64-support*		    wheel*
deps*		      pip*			    wrapper*
destroy-namespace*    pip3*			    yaml-diff*
```

Nuru: The important thing is that I got `vhs` working again. You can update the demo whenever you feel inspired. 

I do not want to install Atmos, because that triggers the Atmos integrations, which will not work in the demo environment (no stacks, no components).

It is kind of ironic that we install about 60 packages and still do not have an easy way to show the power of Geodesic, but in reality, the advantage is we have done all the work to curate, install, and integrate the packages. Plus, as I said before, all the integrations, particularly with Kubernetes (Helm, Helm plugins, Kubeprompt, etc.) and AWS, like assuming roles and connecting to EKS clusters. All hard to show off on a demo running without any code, configuration, or credentials.

PR #974, opened 2025-02-26
Fix Excessive CPU usage, fix --solo

Nuru:
## what

- Fix shell monitoring introduced in v4.1.0 #970 
- Fix `--solo` / `ONE_SHELL=true` option broken in v4.1.0 #970 

## why

- Bug caused shell monitor to burn CPU busy waiting on shell exits rather than sleep and poll
- Oversight in refactoring launch vs exec parameters left out terminal allocation for `--solo` option, causing Geodesic to immediately exit

### DIFF
--- a/rootfs/usr/local/sbin/shell-monitor
+++ b/rootfs/usr/local/sbin/shell-monitor
@@ -18,9 +18,6 @@ shells_are_running() {
 shell_has_exited() {
 	local pid
 
-	trap 'set +x' EXIT
-	set -x
-
 	# If there are no shells left, then they have all exited.
 	[[ "${#wrapper_pids[@]}" -eq 0 ]] && return 0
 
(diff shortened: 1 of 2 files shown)


rootfs/usr/local/sbin/shell-monitor
```diff
@@ -31,25 +36,25 @@ shell_has_exited() {
 # Function to kill all active shell sessions launched by wrapper.
 # This is the shutdown procedure, so we do not care about hogging the CPU.
 kill_shells() {
-	for pid in $(list_wrapper_shells); do
+	for pid in $(list-wrapper-shells); do
```
Benbentwo, line 39: why are we using `-` here instead of `_`?

Benbentwo (dismissed):

rootfs/usr/local/sbin/shell-monitor
Benbentwo: Looks like its another file's function [here](https://github.com/cloudposse/geodesic/blob/main/rootfs/usr/local/sbin/list-wrapper-shells). Personal preference would be update it.

Nuru: Because we can, and by convention we use dashes for command names, which includes functions intended to be called as commands.

